PR #72, opened 2022-02-02
[TITS-5076] Move ESLint, StyleLint and Prettier configs to code-style package

oleg-semyonov:
Key code-style changes:
- Forced function declaration (except for files in `/pages`)
- Enforced consistent `type` imports
- Enforce named exports (not for React Components for now)
- Got rid of `FunctionComponent` type entirely in favor of function declaration and explicitly stating the props type (see https://github.com/facebook/create-react-app/pull/8177 for reasoning)
- Finally got rid of `import React from 'react';`


.eslintrc.js
```diff
@@ -1,185 +1,18 @@
...
-        'no-unused-vars': 'off',
-        '@typescript-eslint/no-unused-vars': [
-            'error',
-            { argsIgnorePattern: '^_', ignoreRestSiblings: true },
-        ],
-
-        'prefer-destructuring': 'warn',
-        'no-nested-ternary': 'warn',
+        'import/no-default-export': 'off',
+        'no-restricted-exports': 'off',
```
oleg-semyonov, line 5: @kudlajz I'm not yet sure about this one. I suppressed the `no-default-export` rule for now, since enforcing it on our components would require quite a big refactor. With named exports we don't really need the `index.ts` files in components (since we're exporting all of them from `@/components/index` anyway)

kudlajz (approved):

PR #78, opened 2022-02-16
[TITS-5133] add changes regarding design 

yaroslavchek:
https://linear.app/prezly/issue/TITS-5133/polishing-bea-designdev-pairing

### DIFF
--- a/components/RichText/Quote.module.scss
+++ b/components/RichText/Quote.module.scss
@@ -22,19 +22,26 @@
 .left {
     display: flex;
     margin-right: $spacing-3;
+    & > svg {
+        color: $color-base-400;
+    }
 }
 
 .right {
     display: flex;
     align-items: flex-end;
     margin-left: $spacing-3;
+    & > svg {
+        color: $color-base-400;
+    }
 }
 
 .content {
-    @include paragraph;
+    @include heading-2;
 
     flex: 1;
     color: $color-base-700;
+    text-align: center;
 }
 
 .alignLeft {
--- a/modules/Layout/Header/Header.tsx
+++ b/modules/Layout/Header/Header.tsx
@@ -125,7 +125,7 @@ function Header({ hasError }: Props) {
                                     [styles.hidden]: isMenuOpen,
                                     [styles.close]: isSearchWidgetShown,
                                 })}
-                                icon={isSearchWidgetShown ? IconClose : IconSearch}
+                                icon={IconSearch}
                                 onClick={toggleSearchWidget}
                                 aria-expanded={isSearchWidgetShown}
                                 aria-controls="search-widget"
--- a/modules/Story/Story.module.scss
+++ b/modules/Story/Story.module.scss
@@ -54,14 +54,14 @@ $story-container-max-width: 926px;
 }
 
 .title {
-    @include heading-1;
+    @include heading-1-1;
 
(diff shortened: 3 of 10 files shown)


modules/Layout/Header/Header.tsx
```diff
@@ -125,7 +125,7 @@ function Header({ hasError }: Props) {
                                     [styles.hidden]: isMenuOpen,
                                     [styles.close]: isSearchWidgetShown,
                                 })}
-                                icon={isSearchWidgetShown ? IconClose : IconSearch}
```
oleg-semyonov, line 128: ❌ 

This was done for the mobile view, and there the icon does close the search. So there should be a more robust solution, in that on desktop it's always Search icon, but on mobile it switches depending on the state. You can use the `useDevice` hook to check if we're on mobile screen or not.

modules/Story/Story.module.scss
```diff
@@ -54,14 +54,14 @@ $story-container-max-width: 926px;
 }
 
 .title {
-    @include heading-1;
+    @include heading-1-1;
```
oleg-semyonov, line 57: This looks like a typo 🤔

styles/variables/_typography.scss
```diff
@@ -9,6 +9,7 @@ $font-size-m: 1.25rem; // 20px
 $font-size-l: 1.5rem; // 24px
 $font-size-xl: 2rem; // 32px
 $font-size-labels: 1rem; // 16px
+$font-size-xll: 2.25rem;
```
oleg-semyonov, line 12: Typo :)
Also it should be moved a line up, so that the order makes a bit more sense.

```suggestion
$font-size-xxl: 2.25rem;
```

styles/mixins/_typography.scss
```diff
@@ -5,6 +5,11 @@
     line-height: $line-height-xl;
     font-weight: $font-weight-bold;
 }
+@mixin heading-1-1 {
```
oleg-semyonov, line 8: ❌ 
Let's name it `heading-1-xxl` so that it's more clear what the mixin does.

Also, please don't forget to separate blocks with empty lines.

oleg-semyonov (changes requested):
I've added some comments that prevent me from hitting approve (blocking comments are marked with ❌ )

modules/Layout/Header/Header.tsx
yaroslavchek: I see, I missed the mobile view point here. I've changed it to isSearchWidgetShown && isMobile ? IconClose : IconSearch

components/RichText/Quote.module.scss
```diff
@@ -22,19 +22,26 @@
 .left {
     display: flex;
     margin-right: $spacing-3;
+    & > svg {
```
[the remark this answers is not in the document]
yaroslavchek: Good to know, thanks.

styles/variables/_typography.scss
yaroslavchek: Done

styles/mixins/_typography.scss
yaroslavchek: done

oleg-semyonov (approved):
Nice! 👍🏻

PR #79, opened 2022-02-16
[TITS-5139] Load Bea theme preset instead of active theme

oleg-semyonov:
+ Add fallbacks if theme preset is not loaded

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN --mount=type=secret,id=NEXT_PUBLIC_HCAPTCHA_SITEKEY \
     export NEXT_PUBLIC_SENTRY_DSN=$(cat /run/secrets/NEXT_PUBLIC_SENTRY_DSN) && \
     export SENTRY_ORG="prezly" && \
     export SENTRY_PROJECT="theme-nextjs-bea" && \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN --mount=type=secret,id=NEXT_PUBLIC_HCAPTCHA_SITEKEY \
     export NEXT_PUBLIC_SENTRY_DSN=$(cat /run/secrets/NEXT_PUBLIC_SENTRY_DSN) && \
     export SENTRY_ORG="prezly" && \
     export SENTRY_PROJECT="theme-nextjs-bea" && \
+    export PREZLY_THEME_UUID="73015107-ac86-418b-9120-4ffa439d5c0f" && \
```
marko-ogg, line 23: Have you checked with Camil about this?

marko-ogg (approved):
👍🏻 I'm so excited to play with previews after deployment of this one 😄

Dockerfile
camilb: @oleg-semyonov Do you need this during the build only? This will be active only during this phase. If you need it at runtime, we need to set it in k8s deployment.

camilb: @oleg-semyonov Done in https://github.com/prezly/eks/pull/28

PR #80, opened 2022-02-16
[MT-4553] Feature - Legit HTML blocks renderer

e1himself:


oleg-semyonov (approved):
Great stuff! Thank you Ivan :D

PR #84, opened 2022-02-23
[TITS-5138] Inject GTM into the page

kudlajz:


modules/analytics/components/Analytics.tsx
```diff
@@ -85,6 +86,27 @@ function Analytics() {
         }
     });
 
+    if (newsroom?.ga_tracking_id) {
+        return (
+            <Head>
+                <script
+                    async
+                    src={`https://www.googletagmanager.com/gtag/js?id=${newsroom.ga_tracking_id}`}
+                />
```
oleg-semyonov, line 95: There is a [Script](https://nextjs.org/docs/basic-features/script) component in Next that can be used here instead :)

oleg-semyonov, line 95: There is a [Script](https://nextjs.org/docs/basic-features/script) component in Next that can be used here instead :)

oleg-semyonov: Also, Gijs mentioned that we support both the new GTag identifers and the older `GA-` IDs. I believe those require a different script, please check with the GA docs

oleg-semyonov (commented):
Great stuff! I have a couple suggestions :)

modules/analytics/components/Analytics.tsx
kudlajz: I've been into Google Analytics dashboard and the tracking code is the same, only the ID changes as far as I can see. I've tested it with both and I've seen some traffic in Google Analytics dashboard but I will check it after deployment as well.

I think you've meant the older `UA-` IDs. The newer ones are `G-`.

oleg-semyonov: Right, my bad :)
Good that it works!

oleg-semyonov (approved):

PR #85, opened 2022-02-28
[TITS-5249] Show date and subtitle only if enabled

kudlajz:


components/StoryCards/StoryCard.module.scss
```diff
@@ -93,22 +101,19 @@ $small-card-image-size-desktop: 164px;
...
         display: grid;
-        grid-template-columns: $small-card-image-size auto;
+        grid-template-columns: $small-card-image-width auto;
         gap: 0 $spacing-4;
         align-items: center;
     }
 
     @include desktop-up {
-        grid-template-columns: $small-card-image-size-desktop auto;
+        grid-template-columns: $small-card-image-width auto;
```
oleg-semyonov, line 110: Was this changed from Hugo's input? 🤔 

Honestly, I don't like how the images are now looking stretched:
image

I believe that square images look much better. I also remember discussing it with Hugo and he agreed that square images are better 
(@iamhugofelix please correct me if I'm wrong or if you changed your mind on that 🙏🏻 )
image

oleg-semyonov (commented):
Great work on the date and subtitle! 👍🏻 

However, I don't want to hit approve just yet, since this PR also includes the card images styling change, which doesn't look good IMO, and isn't truly related to the original task. I'd like to hear an opinion from @iamhugofelix before approving.

components/StoryCards/StoryCard.module.scss
kudlajz: I was just following the Figma file.

I agree that square images look better but I am not sure if it's gonna work because now with categories, title, subtitle and date all being possibly visible there's not enough space.

Just look at the Cern story and how big the title is and it's not even rendering a subtitle.

@iamhugofelix Should we keep using the dimensions that we've had so far?

iamhugofelix: We should use the square image every time it is possible. The only reason for stretching it is for space purposes. Check how it should behave ⬇


️

kudlajz: Okay I've changed the image size back to the original value but used `min-height` and `height: 100%` so that the image stretches vertically if needed.

It now looks like this:
Screenshot 2022-03-02 at 13 39 40

oleg-semyonov: That's perfect, thanks!

oleg-semyonov (approved):

PR #92, opened 2022-03-15
[TITS-5216] Introduce NEXT_PUBLIC_TRACKING_ENABLED variable

kudlajz:


next.config.js
```diff
@@ -43,6 +43,11 @@ const moduleExports = {
             },
         ];
     },
+    env: {
+        NEXT_PUBLIC_IS_TRACKING_ENABLED: JSON.parse(
+            process.env.NEXT_PUBLIC_IS_TRACKING_ENABLED || true,
+        ),
```
kudlajz, line 49: I've thought about it and I think `NEXT_PUBLIC_IS_TRACKING_ENABLED` is a better name to use.

We already have `isTrackingEnabled` prop in our app so this is an extension to that, allowing anyone to disable tracking if they want.

It's also more descriptive than `IS_PREVIEW` and we won't have to introduce yet another variable when we find out we need to disable tracking because of some other use case.

So in this case we would provide `NEXT_PUBLIC_IS_TRACKING_ENABLED=false` for preview environment.

Let me know what you think @oleg-semyonov and @marko-ogg.

oleg-semyonov (commented):
One minor thing to fix before I can hit Approve :)

next.config.js
kudlajz: It works like this because `process.env.NEXT_PUBLIC_IS_TRACKING_ENABLED` is a string. So it's either `true` or `false` string which is always truthy. Only if you don't provide the variable at all or you leave it empty, it's gonna be falsy and thus will default to `true`.

oleg-semyonov: Oh, sorry, I keep forgetting that those are always strings. Then all totally good, approving! :)

PR #93, opened 2022-03-17
[TITS-5262] Properly color boilerplate links

kudlajz:


oleg-semyonov (commented):
The original issue was about headings not having the correct color (h2, h3 tags), but it's nice that you fixed links too :)

Please add a fix for headings and it will be perfect 👍🏻

kudlajz:
Oh yeah, I was wondering why the title says "headings" 😄

modules/Layout/Boilerplate/Boilerplate.module.scss
```diff
@@ -40,13 +40,18 @@
 
 .about {
     margin-bottom: $spacing-5;
+
+    * {
```
kudlajz, line 44: There's a bunch of options (headings 1 to 6, quote etc.) so this felt like the simplest solution that does not affect anything else.

oleg-semyonov: Should be fine, but I would go with `color: inherit`, just so we don't need to maintain it in multiple places.

PR #95, opened 2022-03-18
[TITS-5290] Fix issues with Chinese locales

oleg-semyonov:
The actual fixes to the issue were made in `theme-kit-nextjs`:
- https://github.com/prezly/theme-kit-nextjs/commit/c73e88fdbee599c6aad863fe17442432b64410eb
- https://github.com/prezly/theme-kit-nextjs/commit/b10bafa2e8b1e3b9e31b77e43778a287564f3460

### DIFF
--- a/modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
+++ b/modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
@@ -1,6 +1,5 @@
 import {
     getLanguageDisplayName,
-    getLocaleDirection,
     getUsedLanguages,
     LocaleObject,
     useCurrentLocale,
@@ -10,7 +9,7 @@ import {
     useLanguages,
 } from '@prezly/theme-kit-nextjs';
 import classNames from 'classnames';
-import { useEffect, useMemo } from 'react';
+import { useMemo } from 'react';
 
 import { Dropdown } from '@/components';
 import { IconGlobe } from '@/icons';
@@ -44,27 +43,6 @@ function LanguagesDropdown({ buttonClassName, navigationItemClassName, hasError 
             (language) => language.code !== currentLocale.toUnderscoreCode(),
         );
     }, [currentLocale, languages]);
-
-    // Dynamically update the language attributes on `html` and `meta` tags when changing language.
-    useEffect(() => {
-        const htmlElement = document.querySelector('html');
-
-        if (!htmlElement) {
-            return;
-        }
-
-        // TODO: The direction can be pulled from the Language object
-        const direction = getLocaleDirection(currentLocale);
-        const localeCode = currentLocale.toHyphenCode();
-
-        htmlElement.setAttribute('dir', direction);
-        htmlElement.setAttribute('lang', localeCode);
-
-        const metaTag = document.querySelector('meta[name="og:locale"]');
-        if (metaTag) {
-            metaTag.setAttribute('content', localeCode);
-        }
-    }, [currentLocale]);
 
     // Don't show language selector if there are no other locale to choose
     if (!currentLanguage || displayedLanguages.length < 1) {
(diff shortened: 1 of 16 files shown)


modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
```diff
@@ -44,6 +45,27 @@ function LanguagesDropdown({ buttonClassName, navigationItemClassName, hasError
         );
     }, [currentLocale, languages]);
 
+    // Dynamically update the language attributes on `html` and `meta` tags when changing language.
```
oleg-semyonov, line 48: This is the same fix that was done for Privacy Portal: https://github.com/prezly/privacy-portal/commit/a11027cf1ebce8e9dedf00b09827d4e9753ad052

package.json
```diff
@@ -29,7 +29,7 @@
     "@prezly/content-renderer-react-js": "^0.13.0",
     "@prezly/sdk": "^6.16.0",
     "@prezly/slate-types": "^0.19.0",
-    "@prezly/theme-kit-nextjs": "^1.6.0",
+    "@prezly/theme-kit-nextjs": "^1.6.2",
```
oleg-semyonov, line 32: This is the actual fix for Chinese locales :)

pages/_document.tsx
```diff
@@ -2,6 +2,10 @@ import type { PageProps } from '@prezly/theme-kit-nextjs';
 import { DEFAULT_LOCALE, getLocaleDirection, LocaleObject } from '@prezly/theme-kit-nextjs';
 import Document, { Head, Html, Main, NextScript } from 'next/document';
 
+/**
+ * Note that the attributes set here only apply on the initial page load.
+ * All subsequent locale changes are handled on the client side in `CulturePicker` component.
```
kudlajz, line 7: There is no `CulturePicker` component 🤔

modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
kudlajz: Isn't there a full page refresh when you change culture via the dropdown? Because otherwise I don't see a point in dynamically changing it.

kudlajz (commented):
I'm quite bamboozled 🤔

modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
kudlajz: We also don't support dynamic translations since we've dropped that in favor of server-side loading, right?

pages/_document.tsx
oleg-semyonov: Oh, thanks for the catch, forgot to rename it from the Privacy Portal :)

modules/Layout/Header/LanguagesDropdown/LanguagesDropdown.tsx
oleg-semyonov: > Isn't there a full page refresh when you change culture via the dropdown? Because otherwise I don't see a point in dynamically changing it.

No, the way Next works, after initial page load, all other subsequent page changes are handled on the client.
Translations are updated dynamically, since the server-side props are still requested on page change.

Basically, everything other than `_document` is re-rendering on page changes.

kudlajz: > No, the way Next works, after initial page load, all other subsequent page changes are handled on the client.

I know that's how it works normally but in `LanguagePicker` we are forcing a refresh and I can see in the Network tab that it actually does a full refresh and also the `lang` attribute on `<html>` gets updated.

The language picker works dynamically in privacy portal due to the way it's coded but not in Bea.

kudlajz: So I don't think we need this code at all.

oleg-semyonov: > The language picker works dynamically in privacy portal due to the way it's coded but not in Bea.

Oh, you're right!
So maybe I'd be better off doing the same for Privacy Portal 🤔 
I'll revert the commit then.

oleg-semyonov:
Merging this one, since after code review changes the only thing left was a package version bump :)

PR #97, opened 2022-03-23
Slightly refactored useInfiniteLoading hook

kudlajz:
Since we won't go with extracting the hook at least we can bring the small refactoring. First seen here: https://github.com/prezly/theme-kit-nextjs/pull/18

### DIFF
--- a/modules/InfiniteStories/lib/useInfiniteStoriesLoading.ts
+++ b/modules/InfiniteStories/lib/useInfiniteStoriesLoading.ts
@@ -45,13 +45,8 @@ export function useInfiniteStoriesLoading(
 
     const { canLoadMore, data, isLoading, loadMore, resetData } =
         useInfiniteLoading<StoryWithImage>({
-            fetchingFn: async (nextPage: number) => {
-                const { stories } = await fetchStories(
-                    nextPage,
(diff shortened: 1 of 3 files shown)


modules/InfiniteStories/lib/useInfiniteStoriesLoading.ts
```diff
@@ -45,13 +45,8 @@ export function useInfiniteStoriesLoading(
...
     const { canLoadMore, data, isLoading, loadMore, resetData } =
         useInfiniteLoading<StoryWithImage>({
-            fetchingFn: async (nextPage: number) => {
-                const { stories } = await fetchStories(
-                    nextPage,
-                    pagination.pageSize,
-                    category,
-                    currentLocale,
-                );
+            fetchingFn: async (nextPage: number, pageSize) => {
```
oleg-semyonov, line 48: No need for explicitly stating type here anymore, right?

oleg-semyonov (approved):

modules/InfiniteStories/lib/useInfiniteStoriesLoading.ts
kudlajz: Correct 👍

PR #102, opened 2022-03-29
[TITS-5321] [TITS-5328] Adjust image sizes and titles/subtitles for story cards

oleg-semyonov:


components/StoryCards/HighlightedStoryCard.module.scss
```diff
@@ -84,6 +74,10 @@
     margin-bottom: 0;
 }
 
+.hugeTitle {
```
kudlajz, line 77: Since we're using CSS modules I thought it would be better to use modifier classes, e.g. `&.huge` inside `.title` block.

That way in JSX you can just use `style.title` and `style.huge` and it will be automapped.

oleg-semyonov: Yeah, that makes sense! Thanks for the tip!

PR #103, opened 2022-03-30
[TITS-5330] Use an excerpt from story text as a fallback for story page description

oleg-semyonov:
~~Ported the logic for excerpt description from Marcel to use as fallback in the SEO description of a story page.~~

~~Here's how it works:~~

- ~~It looks for the earliest text nodes in the story (but not later than 5th node, since it wouldn't make sense to use text from middle/end of the story)~~
- ~~It returns the text content of either the first of these nodes, or combines multiple nodes to fit in maximum of 300 characters.~~

We decided to compose the text excerpt in the API instead.

You can check it here: https://theme-nextjs-bea-the-good-newsroom-git-feature-t-90ac5e-prezly1.vercel.app/this-is-a-test-story


components/StorySeo/lib/getStoryExcerpt.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (node.children as Node[]).reduce(
+        (total: number, child) => total + getNodeTextLength(child),
+        0,
+    );
+}
+
+const MAX_NODE_INDEX_FOR_TEXT_NODES = 5;
+const MAX_TOTAL_TEXT_LENGTH = 300;
+
+export function getStoryExcerpt(story: ExtendedStory): string {
```
kudlajz, line 29: Works nice, good job 👍 

I am not sure if this should be part of the theme code 🤔 I think it would be better to have a function like this in one of the `@prezly/slate` packages. IMHO themes should not care about the story content structure - that's why we have the `@prezly/content-renderer-react-js`. Separation of concerns.

Also, now you'll have to duplicate this function in each theme unless you extract it into the theme kit, in which case you'll have to add `@prezly/content-renderer-react-js` and all the related packages as dependencies there.

The code for old editor stories can go to theme kit directly IMHO.

kudlajz (commented):
Works nice, good job 👍 Although I've left a comment explaining why I think this is not a good idea to have it here.

components/StorySeo/lib/getStoryExcerpt.tsx
oleg-semyonov: That's a good point Lukas! This is exactly the question I asked in Linear :)

> Also, now you'll have to duplicate this function in each theme unless you extract it into the theme kit, in which case you'll have to add @prezly/content-renderer-react-js and all the related packages as dependencies there.

This one didn't come to my mind 🤔 
It is right, we don't want to bring extra deps to `theme-kit-nextjs` just for this method.

But I'm not entirely sure if Slate packages are a good place to store this function, too.
It's kinda in between the concerns of a theme and the Slate renderer.

Would be cool if @e1himself could give his opinion on that, because he's basically the owner of Slate repos right now.

oleg-semyonov: We decided that even better option would be to expose the `summary` from the API.

PR #107, opened 2022-04-06
Add support for Placeholder nodes

oleg-semyonov:


components/SlateRenderer/components/Placeholder.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { StoryPublicationDate } from '@/components';
+
+interface Props {
+    node: PlaceholderNode;
+}
+
+function Placeholder({ node }: Props) {
+    const currentStory = useCurrentStory();
+
+    if (!currentStory) {
```
OskarGasiorowski, line 13: Just wondering, why `currentStory` could be null?

OskarGasiorowski (approved):

components/SlateRenderer/components/Placeholder.tsx
kudlajz: `currentStory` is defined on `NewsroomContext` but is not always provided except for `/[slug]` and/or the preview `/s/[uuid]` pages.

kudlajz (approved):

PR #109, opened 2022-04-06
[TF-588] Update quote text

fgyimah:


components/RichText/Quote.module.scss
```diff
@@ -4,13 +4,13 @@
     display: flex;
     justify-content: center;
     margin: $spacing-6 auto;
-    padding: $spacing-4;
-    background: $color-base-100;
+    padding-top: $spacing-4;
+    padding-bottom: $spacing-4;
```
oleg-semyonov, line 8: You can use the short syntax here:
```suggestion
    padding: $spacing-4 0;
```

components/RichText/Quote.module.scss
```diff
@@ -33,10 +33,8 @@
 }
 
 .content {
-    @include heading-2;
-
+    font-style: italic;
     flex: 1;
-    color: $color-base-700;
     text-align: center;
```
oleg-semyonov, line 38: Please test with the editor to see that the text in the quote indeed follows the text alignment settings.

components/RichText/Quote.tsx
```diff
@@ -25,7 +25,7 @@ export function Quote({ node, children }: PropsWithChildren<Props>) {
                     [styles.alignRight]: node.align === Alignment.RIGHT,
                 })}
             >
-                {children}
+                <p>{children}</p>
```
oleg-semyonov, line 28: Was there a particular reason for this paragraph tag to be added? I get extra margin on top and bottom from the default browser stylesheet on Chrome (Mac). If this is what you intended, it is better to add these styles explicitly, since we don't want to rely on user agent stylesheets.

oleg-semyonov (commented):
I got a couple of questions, can't approve yet.

components/RichText/Quote.tsx
fgyimah: that was to apply the p/paragraph styles, will change to explicitly use the paragraph class

components/RichText/Quote.module.scss
fgyimah: texted all alignment options and it follows the proposed design inside figma

oleg-semyonov (approved):
Great stuff! 👍🏻

PR #114, opened 2022-04-11
[TF-651] Fix the search box placeholder lengthy issue - make it compact

zourdyzou:
- [x] Still testing and checking it in all languages listed in the newsroom. (CHECKED)


modules/Layout/Header/SearchWidget/components/SearchBar.tsx
```diff
@@ -37,30 +36,7 @@ function SearchBar({ currentRefinement, refine }: Props) {
...
-                                                            {...translations.search.action}
-                                                        />
-                                                        &quot;
-                                                    </>
-                                                ),
-                                            }}
-                                        />
-                                    </span>
-                                ),
+                                inputHintExtra: <span>&nbsp;</span>,
```
zourdyzou, line 39: Apparently, we still need to specify this. 

We could actually do something like this: 
```js
inputHintExtra: <>&nbsp;</>
```

oleg-semyonov: It can be `undefined` as well.
But ideally we would want to remove the string from `themes-intl-messages`, so that this parameter is not needed anymore.

oleg-semyonov (changes requested):
Looks good, but it would be nice to make the solution more complete, see my comment.

modules/Layout/Header/SearchWidget/components/SearchBar.tsx
zourdyzou: So a proper thing to do would be either to delete the {inputHintExtra} from `themes-intl-messages` or to delete the 
`search.inputHintExtra` in the crowdin editor?

modules/Layout/Header/SearchWidget/components/SearchBar.tsx
```diff
@@ -5,7 +5,6 @@ import { connectSearchBox } from 'react-instantsearch-dom';
 import { FormattedMessage, useIntl } from 'react-intl';
 
 import { Button, FormInput } from '@/components';
-import { IconEnterKey } from '@/icons';
```
oleg-semyonov, line 8: Please remove the icon completely (the svg file and the definition) if it's not used anymore.

modules/Layout/Header/SearchWidget/components/SearchBar.module.scss
```diff
@@ -47,19 +47,9 @@
...
-    }
-}
-
-.inputHintDesktop {
-    @include mobile-only {
-        display: none;
-    }
 }
 
 .keyHint {
```
oleg-semyonov, line 55: This classname is not used anymore and should be removed.

package.json
```diff
@@ -31,7 +31,7 @@
     "@prezly/sdk": "^6.21.0",
     "@prezly/slate-types": "^0.19.0",
     "@prezly/theme-kit-nextjs": "^1.6.2",
-    "@prezly/themes-intl-messages": "^1.9.1",
+    "@prezly/themes-intl-messages": "^1.9.2",
```
oleg-semyonov, line 34: Please update to `v1.10.0`

oleg-semyonov (changes requested):

oleg-semyonov (approved):

PR #117, opened 2022-04-13
[TF-635] Handle API errors better

oleg-semyonov:


pages/_app.tsx
```diff
@@ -18,6 +18,14 @@ function App({ Component, pageProps }: AppProps) {
     const { localeCode, newsroom, currentStory } = newsroomContextProps;
     const locale = useMemo(() => LocaleObject.fromAnyCode(localeCode), [localeCode]);
 
+    // `newsroomContextProps` can be undefined, if there was error when fetching the newsroom props.
+    // This can happen due to connection issues, or incorrect credentials in your .env file.
+    // In this case, a 500 error page would be rendered, which shouldn't rely on the Newsroom Context (especially when statically generated).
+    if (!newsroomContextProps) {
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        return <Component {...customPageProps} />;
+    }
```
kudlajz, line 27: I just realised this part needs to be defined BEFORE the
```
const { localeCode, newsroom, currentStory } = newsroomContextProps;
```

because the code crashes on that line.

@oleg-semyonov

oleg-semyonov: Oh snap! Thanks for the catch Lukas! I'll push a hot fix for that

oleg-semyonov: Done: https://github.com/prezly/theme-nextjs-bea/commit/563a204e5f6643e032266d946b7c3a6ce58a3887

PR #124, opened 2022-04-21
Add story bookmark renderer

aspirisen:


components/SlateRenderer/SlateRenderer.tsx
```diff
@@ -48,7 +49,7 @@ function SlateRenderer({ nodes }: Props) {
 
     return (
         <div className={styles.renderer}>
-            <Renderer nodes={nodes}>
+            <Renderer nodes={nodes} defaultComponents>
```
aspirisen, line 52: @e1himself i assume that we need to add default components?

e1himself: It's not necessary, as `defaultComponents` is true by default.

e1himself (approved):
Looks good :+1:

PR #127, opened 2022-04-26
Dev env

aspirisen:


pages/_app.tsx
```diff
@@ -10,6 +10,7 @@ import type { BasePageProps } from 'types';
 import '@prezly/uploadcare-image/build/styles.css';
 import 'modern-normalize/modern-normalize.css';
 import '../styles/styles.globals.scss';
+import '@prezly/content-renderer-react-js/styles.css';
```
aspirisen, line 13: I have build error from next js - i must have global styles only in `_app.tsx`

.nvmrc
```diff
@@ -1 +1 @@
-v17.1.0
+v16.14.2
```
aspirisen, line 1: I just wanted to align the node version, let me know if this is an issue

oleg-semyonov: We have an issue currently with NextJS themes kubernetes pods crashing on Node 16, so we rolled back the deployment version to v14. But for local development Node 16 should be fine I believe.

pages/_app.tsx
oleg-semyonov: It's better to move this import a bit higher up, right before `uploadcare-image` styles

oleg-semyonov (approved):
One small comment on import order, otherwise good! Glad that not much needed to change :)

.github/workflows/lint.yml
```diff
@@ -11,6 +11,11 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v2
 
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
```
e1himself, line 17: Oh, nice! :+1:

PR #128, opened 2022-04-29
Fix extraneous dependencies

aspirisen:
When i run typecheck found extraneous dependency


package.json
```diff
@@ -40,6 +40,7 @@
     "next": "^12.0.10",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "react-instantsearch-core": "^6.24.0",
```
oleg-semyonov, line 43: We only import types from this package, which is a dependency of `react-instantsearch-dom`. I guess it can be moved to devDeps, or maybe even just `@types/react-instantsearch-core` is enough.

PR #130, opened 2022-05-09
[DEV-5526] Update quotes UI

fgyimah:
How it looks on the various alignments

Screenshot 2022-05-09 at 1 54 09 PM
Screenshot 2022-05-09 at 1 55 13 PM
Screenshot 2022-05-09 at 1 55 29 PM

### DIFF
--- a/components/RichText/Quote.module.scss
+++ b/components/RichText/Quote.module.scss
@@ -1,22 +1,8 @@
 .container {
     @include border-radius-m;
-
-    display: flex;
     margin: $spacing-6 auto;
     padding: $spacing-4;
     background: #f9fafb;
-
-    &.left {
-        justify-content: left;
-    }
-
-    &.center {
-        justify-content: center;
-    }
-
-    &.right {
-        justify-content: right;
-    }
 }
 
 .left,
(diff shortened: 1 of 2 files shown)


components/RichText/Quote.tsx
```diff
@@ -3,34 +3,36 @@ import { Alignment } from '@prezly/slate-types';
...
 export function Quote({ node, children }: PropsWithChildren<Props>) {
+    const alignment = node.align ?? Alignment.LEFT;
+
     return (
-        <blockquote className={styles.container}>
-            <div className={styles.left} aria-hidden="true">
-                <IconQuoteOpening className={styles.quote} />
-                <IconQuoteOpening className={styles.quote} />
-            </div>
+        <blockquote
```
oleg-semyonov, line 16: ✖️
Same comment here as in Marcel: it can be made much simpler.

Since this block also has a background, a simple border probably won't work, but you can add it on the inner `div` with the text and achieve the same result.

components/RichText/Quote.module.scss
```diff
@@ -2,39 +2,41 @@
     @include border-radius-m;
 
     display: flex;
-    justify-content: center;
     margin: $spacing-6 auto;
-    padding: $spacing-4 0;
-}
+    padding: $spacing-4;
+    background: #f9fafb;
```
oleg-semyonov, line 7: ✖️ 
This should be extracted into a variable, or maybe even aligned with existing shade of gray. Please consult with Hugo on that one.

oleg-semyonov (changes requested):
Same thing as in Marcel - it can be made simpler :)

PR #132, opened 2022-05-11
[TF-605] Back to top button

fgyimah:


modules/Layout/Layout.tsx
```diff
@@ -48,6 +49,32 @@ function Layout({ children, description, imageUrl, title, hasError }: PropsWithC
...
+        };
+    }, []);
+
+    function scrollToTop() {
+        window.scrollTo({
+            top: 0,
+            behavior: 'smooth',
+        });
+    }
+
```
oleg-semyonov, line 77: IMO all of this logic can be moved into the `ScrollToTopButton` component. It doesn't depend on any state or props in the `Layout` component, so it would be nice to encapsulate everything in one component. Same goes for the `isScrollToTopVisible` state

components/ScrollToTopButton/ScrollToTopButton.module.scss
```diff
@@ -0,0 +1,23 @@
+.button {
```
oleg-semyonov, line 1: The way the button is placed is not really optimal. This is especially obvious on mobile:
image

The sizing is also not good - the icon should be a bit bigger IMO, and the spacing is not even, which makes the icon appear not centered vertically.

Here's my take on the button placement:
- The clickable size should be a bit bigger, so that's it easy to click/tap (increasing icon size should help, but maybe it's worth to increase the padding up a notch
- The bottom and right offset should be smaller, at least on smaller screens
- On large screens, it would be nice to place the button closer to the content, e.g. place it in a virtual "container".

modules/Layout/Layout.tsx
```diff
@@ -48,6 +49,32 @@ function Layout({ children, description, imageUrl, title, hasError }: PropsWithC
...
+            } else {
+                setIsScrollToTopVisible(false);
+            }
+        }
+        if (typeof window !== 'undefined') {
+            window.onscroll = scrollListener;
+        }
+
+        return () => {
+            if (window !== null) {
```
oleg-semyonov, line 65: Why is this check different from the one on line 60? 🤔 
It makes sense to just check for `typeof` in both cases.
Also, the way Next works, `useEffect` should only fire on client, so `window` should always be defined. But good to check nevertheless.

oleg-semyonov (changes requested):
The solution is good, but it could use some improvements :)
See my comments.

components/ScrollToTopButton/ScrollToTopButton.tsx
```diff
@@ -1,22 +1,46 @@
...
+function ScrollToTopButton() {
+    const [isScrollToTopVisible, setIsScrollToTopVisible] = useState(false);
+
+    useEffect(() => {
+        function scrollListener() {
+            if (document.body.scrollTop > 300 || document.documentElement.scrollTop > 300) {
+                setIsScrollToTopVisible(true);
+            } else {
+                setIsScrollToTopVisible(false);
+            }
```
oleg-semyonov, line 19: This can be made simpler:
```suggestion
                setIsScrollToTopVisible(document.body.scrollTop > 300 || document.documentElement.scrollTop > 300);
```

oleg-semyonov: Also better to move the `300` value to a constant

oleg-semyonov (changes requested):
Almost there! Some code-styles suggestions from me

components/ScrollToTopButton/ScrollToTopButton.tsx
fgyimah: updated!

oleg-semyonov (approved):
👍🏻

PR #134, opened 2022-05-13
[TF-602] Bea polishments

fgyimah:


modules/Search/components/Facet.tsx
```diff
@@ -71,7 +74,12 @@ function Facet({ attribute, items, refine }: RefinementListProvided & Refinement
                     <Disclosure.Panel className={styles.panel}>
                         <ul className={styles.list}>
                             {visibleItems.map((item) => (
-                                <li key={`${attribute}_${item.label}`} className={styles.listItem}>
+                                // eslint-disable-next-line
```
oleg-semyonov, line 77: ✖️ 
Why is eslint disabled completely here? 🤔

modules/Search/components/Facet.tsx
```diff
@@ -71,7 +74,12 @@ function Facet({ attribute, items, refine }: RefinementListProvided & Refinement
...
+                                <li
+                                    key={`${attribute}_${item.label}`}
+                                    className={styles.listItem}
+                                    onClick={() => refine(item.value)} // make entire list item clickable
```
oleg-semyonov, line 81: ✖️ 
If you wanted to make the entire element clickable, it's better to not add the `onClick` listener, but rather update styles so that `listItem` doesn't have any padding, and all styles should come from `listItemInner`. That way you will make the whole element clickable with just CSS

oleg-semyonov (changes requested):

modules/Search/components/Facet.tsx
fgyimah: Sure, making the update!

modules/Search/components/Facet.tsx
fgyimah: Updated!

modules/Search/components/Facet.tsx
fgyimah: Updated!

oleg-semyonov (approved):
Nice! 👍🏻

PR #232, opened 2022-07-12
[TF-939] Bea/Greta Bug: Remove hamburguer icon when there is no content in menu

MohammadxAli:
As mentioned in the ticket, when a user doesn't use categories or languages we should hide the mobile menu toggler.
One more case to notice is when the user has a gallery but no images in it, so I tried to handle that case as well.

### DIFF
--- a/hooks/index.ts
+++ b/hooks/index.ts
@@ -1,3 +1,4 @@
 export * from './useDevice';
 export * from './useIsMounted';
 export * from './useThemeSettings';
+export * from './useDisplayedLanguages';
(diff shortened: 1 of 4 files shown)


MohammadxAli:
Hey, @pestaa I added you as reviewer since Oleg is away, feel free to review it when you have time.

Oh btw, the lint fixes commits were due to setting up the project correctly on my side, running `npm run lint` gave me a bloated output of the CRLF/LF issues so I couldn't really check the actual problem without pushing to branch.

It took me a while to figure it out but I'm all set now :) Let me know if you want me to change anything.

MohammadxAli:
@pestaa Sure! Just need to figure out how to do it :) Didn't have experience using rebase effectively before. I'll let you know if any questions come up over Slack.

PR #239, opened 2022-07-15
[TF-945] Bea improvements: Smaller Story Cards

MohammadxAli:
- [x] The card images aspect ratio changed to 135:92 (same as main card)
- [x] Remove subtitles for this card type
- [x] “Title section” bottom/top-padding removed
- [x] Date color changed to $neutral-600
- [x] New truncation behavior

### DIFF
--- a/components/StoryCards/StoryCard.tsx
+++ b/components/StoryCards/StoryCard.tsx
@@ -2,7 +2,7 @@ import { StoryPublicationDate } from '@prezly/themes-ui-components';
 import classNames from 'classnames';
 import Link from 'next/link';
 
-import { useThemeSettings } from '@/hooks';
+import { useDevice, useThemeSettings } from '@/hooks';
 import type { StoryWithImage } from 'types';
 
 import CategoriesList from '../CategoriesList';
@@ -18,9 +18,10 @@ type Props = {
 function StoryCard({ story, size = 'small' }: Props) {
     const { categories, title, subtitle } = story;
     const { showDate, showSubtitle } = useThemeSettings();
-
+    const { isTablet } = useDevice();
+    const hasCategories = categories.length > 0;
     const HeadingTag = size === 'small' ? 'h3' : 'h2';
-
+    const shouldShowSubtitle = isTablet ? true : size !== 'small';
     return (
         <div
             className={classNames(styles.container, {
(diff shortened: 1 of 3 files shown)


components/StoryCards/StoryCard.tsx
```diff
@@ -18,9 +18,10 @@ type Props = {
 function StoryCard({ story, size = 'small' }: Props) {
     const { categories, title, subtitle } = story;
     const { showDate, showSubtitle } = useThemeSettings();
-
+    const { isTablet } = useDevice();
+    const hasCategories = categories.length > 0;
     const HeadingTag = size === 'small' ? 'h3' : 'h2';
-
+    const shouldShowSubtitle = isTablet ? true : size !== 'small';
```
fgyimah, line 24: this would show subtitles for small cards that are in tablet aspect ratio or lower, I think we should remove for all small cards?

components/StoryCards/StoryCard.tsx
```diff
@@ -48,13 +49,21 @@ function StoryCard({ story, size = 'small' }: Props) {
...
                     </div>
                 )}
-                <HeadingTag className={styles.title}>
+
+                <HeadingTag
+                    className={classNames(
+                        styles.title,
+                        !hasCategories && styles.noCategories,
+                        !showDate && styles.noDate,
+                        !hasCategories && !showDate && styles.noDateAndCategories,
```
fgyimah, line 58: Should we use this style instead to keep it consistent with other parts of the codebase

```suggestion
                       {
                          [styles.noCategories]: !hasCategories,
                          [styles.noDate]: !showDate,
                          [styles.noDateAndCategories]: !hasCategories && !showDate
                       }
```

fgyimah (changes requested):

components/StoryCards/StoryCard.tsx
MohammadxAli: Good point 👍

components/StoryCards/StoryCard.tsx
MohammadxAli: I guess we need to keep subtitles on mobile and tablet devices as their layout supports it, I'll ping Hugo and ask him for more details.

fgyimah (approved):
👍🏾

PR #251, opened 2022-07-19
[TF-948] Use only one type of story card in tablet screens

MohammadxAli:
- Decreased gutter/spacing in tablet to `32px` [according to new design](https://www.figma.com/file/j0ty2jW7wHD4bFcOyWn39a/Working-%F0%9F%9A%A7-Bea-Theme?node-id=745%3A27072)
- Use one type of story card on tablet screens (The vertical variant)
- Adjust spacings to make it closer to the design
- Set the same aspect ratio for images in all types of story cards

I guess it makes more sense to follow `vertical` and `horizontal` variants for story cards instead of `big` | `medium` | `small` sizes. However, that refactor may change lots of stuff but that's something we can consider to simplify CSS.

### DIFF
--- a/components/StoryCards/StoryCard.module.scss
+++ b/components/StoryCards/StoryCard.module.scss
@@ -3,10 +3,6 @@ $small-card-image-size: 230px;
 .container {
     position: relative;
 
-    &:not(:last-of-type) {
-        margin-bottom: $spacing-6;
-    }
-
     /* stylelint-disable-next-line order/order */
     @include desktop-up {
         &:not(:last-of-type) {
@@ -54,7 +50,11 @@ $small-card-image-size: 230px;
 }
 
 .content {
-    padding: $spacing-5 0;
+    padding-top: $spacing-4 - $spacing-half; // 20px
+
+    @include desktop-up {
+        padding: ($spacing-4 - $spacing-half) 0; // 20px 0
+    }
 }
 
 .categories {
@@ -110,7 +110,7 @@ $small-card-image-size: 230px;
 }
 
 .small {
-    @include tablet-up {
+    @include desktop-up {
         display: grid;
         grid-template-columns: $small-card-image-size auto;
         gap: 0 $spacing-4;
@@ -122,13 +122,13 @@ $small-card-image-size: 230px;
     }
 
     .content {
-        @include tablet-up {
+        @include desktop-up {
             padding: 0;
         }
     }
 
     .title {
-        @include tablet-up {
+        @include desktop-up {
             @include heading-3;
             @include ensure-max-text-height(3, $line-height-m);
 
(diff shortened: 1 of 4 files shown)


MohammadxAli:
This PR also resolves [TF-944](https://linear.app/prezly/issue/TF-944) and [TF-943](https://linear.app/prezly/issue/TF-943)

components/StoryCards/StoryCard.module.scss
```diff
@@ -1,12 +1,9 @@
 $small-card-image-size: 230px;
+$spacing-20px: $spacing-4 - $spacing-half; // 1.25rem = 20px That's the only hack I can think of which doesn't mess with other spacings
```
oleg-semyonov, line 2: This looks like a design issue to me 🤔 
It will work for now, but would be nice to figure out together with Hugo or Al to reconsider the spacing system.

cc @iamhugofelix

oleg-semyonov: Also, I think it makes sense to rename it to `$spacing-card`, cause we might change the size later and 20px naming will be not relevant anymore.

components/StoryCards/StoryCard.module.scss
```diff
@@ -161,20 +162,6 @@ $small-card-image-size: 230px;
     }
 }
 
-.medium,
-.big {
```
oleg-semyonov, line 165: > I guess it makes more sense to follow vertical and horizontal variants for story cards instead of big | medium | small sizes. However, that refactor may change lots of stuff but that's something we can consider to simplify CSS.

Yes, it would be really nice! Can we do this in scope of this PR?

modules/InfiniteStories/StoriesList.module.scss
```diff
@@ -7,6 +7,18 @@
...
+        display: grid;
+        grid-template-columns: repeat(1, 1fr);
+        row-gap: $spacing-6;
+    }
+
+    @include tablet-only {
+        display: grid;
+        grid-template-columns: repeat(2, 1fr);
+        gap: $spacing-6 $grid-gutter-tablet;
+    }
```
oleg-semyonov, line 20: The `display` property is the same for all three breakpoints, so it makes sense to move it outside to the root of the selector. The `gap` can also be re-used for mobile and tablet, since mobile only has one column.
And it makes no sense to use `repeat` for mobile where we only have one column, it can be just be `grid-template-columns: 1fr`

components/StoryCards/StoryCard.module.scss
MohammadxAli: As discussed, let's do this in a separate PR.

MohammadxAli:
@oleg-semyonov Refactored styles based on the review, and as discussed let's do the new truncation behavior in a separate PR.

modules/InfiniteStories/StoriesList.module.scss
```diff
@@ -7,20 +7,18 @@
...
 
 .storiesContainer {
+    display: grid;
+    gap: $spacing-6 $grid-gutter-tablet;
+
     @include mobile-only {
-        display: grid;
-        grid-template-columns: repeat(1, 1fr);
-        row-gap: $spacing-6;
+        grid-template-columns: 1fr;
```
oleg-semyonov, line 14: Sorry, missed that part in previous review:
Let's move this line to the root and use `tablet-up` mixin instead of `tablet-only` to make things more streamlined.

oleg-semyonov (approved):
All good! Please address my small code-style comment before merging.

Also, feel free to mark my comments as resolved once you push fix for them :)

PR #256, opened 2022-07-19
[TF-952] Bea/Greta improvements: Story page

fgyimah:


modules/Story/Story.tsx
```diff
@@ -40,35 +40,34 @@ function Story({ story }: Props) {
...
+                        </p>
+                    )}
+                    <StoryLinks url={links.short || links.newsroom_view} />
+                    {headerImage && (
+                        <Image
+                            alt=""
+                            className={classNames({
+                                [styles.mainImage]: !IS_FANCY_IMAGE_ENABLED,
+                                [styles.fullWidthImage]: IS_FANCY_IMAGE_ENABLED,
+                            })}
```
oleg-semyonov, line 64: I guess we're not going with the full width image option after all, so these conditional styles can probably be removed (along with the `fullWidthImage` class in the stylesheet and the constant on top).

@iamhugofelix let us know if you think this might be used later down the road.

modules/Story/Story.tsx
```diff
@@ -40,35 +40,34 @@ function Story({ story }: Props) {
...
-                    </div>
+                    {hasCategories && <CategoriesList categories={categories} showAllCategories />}
                     <h1 className={styles.title}>{title}</h1>
                     <p className={styles.subtitle}>{subtitle}</p>
+                    {showDate && (
+                        <p className={styles.date}>
+                            <StoryPublicationDate story={story} />
+                        </p>
+                    )}
+                    <StoryLinks url={links.short || links.newsroom_view} />
```
oleg-semyonov, line 57: ❌ 
The popup shown after clicking the Copy URL button looks weird:

Screen Shot 2022-07-20 at 12 49 34

modules/Gallery/Gallery.tsx
```diff
@@ -1,7 +1,9 @@
 import type { NewsroomGallery } from '@prezly/sdk';
 import { getAssetsUrl, getUploadcareGroupUrl } from '@prezly/theme-kit-nextjs';
+import { useEffect, useState } from 'react';
 
 import { SlateRenderer } from '@/components';
+import StoryLinks from '@/components/StoryLinks';
```
oleg-semyonov, line 6: Let's add `StoryLinks` component to the index, so that we can import it from `@/components` as every other component in the theme.

modules/Story/Story.tsx
```diff
@@ -6,7 +6,6 @@ import Image from '@prezly/uploadcare-image';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
 
-import { StoryStickyBar } from '@/components';
```
oleg-semyonov, line 9: If this component is no longer used, let's remove it.

oleg-semyonov (changes requested):
Looks good! Found a visual bug though. Also couple things to clean up :)

modules/Story/Story.tsx
fgyimah: Removed

modules/Gallery/Gallery.tsx
fgyimah: Updated

modules/Story/Story.tsx
fgyimah: Updated the design

oleg-semyonov (changes requested):
Almost there :)

PR #257, opened 2022-07-20
[TF-953] Bea Bug: Weird text formatting on Story

fgyimah:
this targets blockquotes inside htmlContent

### DIFF
--- a/components/RichText/styles.module.scss
+++ b/components/RichText/styles.module.scss
@@ -120,7 +120,6 @@
     border-color: inherit;
 }
 
-// add quote support to html content
 .htmlContent blockquote {
     @include border-radius-m;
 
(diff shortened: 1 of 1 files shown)


components/RichText/styles.module.scss
```diff
@@ -119,3 +119,18 @@
     border-collapse: collapse;
     border-color: inherit;
 }
+
+// add quote support to html content
```
oleg-semyonov, line 123: I think this comment is redundant :)

oleg-semyonov (approved):

components/RichText/styles.module.scss
fgyimah: Sure, taking it out! :)

PR #260, opened 2022-07-20
[TF-1013] Bea/Greta improvements: Media Gallery

fgyimah:


components/GalleryCard/GalleryCard.module.scss
```diff
@@ -11,7 +11,7 @@
 .thumbnail {
     @include border-radius-s;
 
-    height: 340px !important;
+    height: 236px !important;
```
oleg-semyonov, line 14: Shouldn't it use the same aspect-ratio mixin as the story cards? I guess the idea was to make all card images have the same ratio?

components/PageTitle/PageTitle.tsx
```diff
@@ -7,11 +7,12 @@ interface Props {
     title: string;
     subtitle?: ReactNode;
     className?: string;
+    bgClassName?: string;
```
oleg-semyonov, line 10: Let's try to not shorten the class names like this. `backgroundClassName` is not that long after all :)

oleg-semyonov: Also, I was under the impression, that we want to remove the bg color and border from ALL Page Title blocks. Then this extra class name will not be needed. Please correct me if I'm mistaken.

modules/Galleries/GalleriesList.module.scss
```diff
@@ -1,5 +1,5 @@
 .container {
-    padding: $spacing-8 0 0;
+    padding: 0;
```
oleg-semyonov, line 2: You can remove this class entirely, since the block will not have padding set either way.

oleg-semyonov (changes requested):
Looks good!
Although some improvements could be made under the hood

components/GalleryCard/GalleryCard.module.scss
fgyimah: Yeah will update, will have same as the medium story card

components/PageTitle/PageTitle.tsx
fgyimah: Checked the other pages and all page titles are having this same design now, will update and remove the extra classes

components/PageTitle/PageTitle.module.scss
```diff
@@ -1,18 +1,14 @@
 .background {
     @include full-width;
```
oleg-semyonov, line 2: Since we don't have a background anymore, there is not point in having this container at all. So you can safely remove the class and the `div` in the component.

components/PageTitle/PageTitle.module.scss
```diff
@@ -1,18 +1,14 @@
...
-
-    background: $color-base-50;
-    border-bottom: 1px solid $color-borders;
 }
 
 .container {
     @include container;
```
oleg-semyonov, line 6: Same here, since the block doesn't need to escape the main container anymore, we don't need the container mixin. So you can leave just the vertical padding in the styles.

modules/Galleries/GalleriesList.module.scss
```diff
@@ -1,5 +1,5 @@
 .container {
-    padding: 0;
+    padding: $spacing-8 0 0;
```
oleg-semyonov, line 2: This spacing is not consistent with the spacing on the Category page. I think the spacing for categoryList should be updated in the `InfiniteStories` component

oleg-semyonov (changes requested):
Spotted a couple more things to improve

components/PageTitle/PageTitle.module.scss
fgyimah: updated

components/PageTitle/PageTitle.module.scss
fgyimah: updated

modules/Galleries/GalleriesList.module.scss
fgyimah: updated

components/PageTitle/PageTitle.module.scss
```diff
@@ -1,10 +1,4 @@
-.background {
-    @include full-width;
-}
-
 .container {
-    @include container;
-
     padding: $spacing-8 0 0;
 
     @include mobile-only {
```
oleg-semyonov, line 4: This looks wrong:
image

Let's simplify both padding rules on lines 2 and 5 to only specify `padding-top`.

oleg-semyonov (changes requested):
One tiny bug left to fix :)

components/PageTitle/PageTitle.module.scss
fgyimah: Sure, on it

oleg-semyonov (approved):
👍🏻

PR #262, opened 2022-07-21
[TF-1016] Bea/Great Bug: Comma after the category is wrong

fgyimah:


components/CategoriesList/CategoriesList.tsx
```diff
@@ -52,7 +52,7 @@ function CategoriesList({ categories, showAllCategories = false, isStatic }: Pro
     }, [categories, showExtraCategories, currentLocale]);
 
     return (
-        <>
+        <div>
```
oleg-semyonov, line 55: I see that the problem was with the `:last-child` selector.

But, adding this `div` effectively disabled the styles applied from the parent containers in story cards.
It would make sense to move the base layout styles from wrapping container in the story cards to this div, and add a `className` prop to the component, so that we can add extra styles.

So, for `HighlightedStoryCard`:
- Remove the `categoriesWrapper` div and its class
- The font-size and line-height properties are redundant and can be removed
- The flex properties should move to the `CategoriesList` styles

For `StoryCard`:
- Remove the `categories` div
- Apply the `categories` from the component stylesheet as `className` prop for Categories list
- Remove flex properties from `categories` class name.

oleg-semyonov (changes requested):

components/CategoriesList/CategoriesList.tsx
fgyimah: Applied these changes

components/StoryCards/HighlightedStoryCard.tsx
```diff
@@ -36,18 +36,16 @@ function HighlightedStoryCard({ story }: Props) {
                 </a>
             </Link>
             <div className={styles.content}>
-                <div className={styles.categoriesWrapper}>
-                    {/* 
+                {/* 
                         Since the position of date is changed the separator is no longer needed.
                         Commenting out instead of deleting in case we need to revert back.
```
oleg-semyonov, line 41: I think these commented blocks can be safely deleted :)

oleg-semyonov (approved):
Nice!

components/StoryCards/HighlightedStoryCard.tsx
fgyimah: removed!

PR #276, opened 2022-07-26
[TF-951] Bea/Greta improvements: Search + Results page

fgyimah:


modules/Layout/Header/SearchWidget/components/CategoriesList.tsx
```diff
@@ -44,23 +42,12 @@ function CategoriesList({ filteredCategories }: Props) {
...
                     {showAllCategories ? (
                         <FormattedMessage {...translations.search.viewLess} />
                     ) : (
                         <FormattedMessage {...translations.search.viewMore} />
                     )}
-                    <IconSearchCaret
-                        width={12}
-                        height={12}
-                        className={classNames(styles.caret, {
-                            [styles.caretOpen]: showAllCategories,
```
oleg-semyonov, line 61: This class name should also be removed from the stylesheet

modules/Search/components/Subtitle.tsx
```diff
@@ -0,0 +1,31 @@
...
+                    values={{
+                        resultsCount: <b>{resultsCount}</b>,
+                        searchQuery: (
+                            <>
+                                &quot;<b>{searchQuery}</b>&quot;
+                            </>
+                        ),
+                    }}
+                />
+            ) : undefined}
```
oleg-semyonov, line 26: Currently, the content shifts when this text is updated.
If you want to prevent it, you need to set a height on the parent `p` element.
Otherwise, you can simplify the code by just returning `null` earlier when there is no search query

modules/Search/components/Facet.module.scss
```diff
@@ -1,27 +1,53 @@
...
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        width: 100%;
     }
 }
 
-.header {
-    @include text-label;
+.menu {
```
oleg-semyonov, line 44: ❌ 
I find the horizontal spacing on the dropdown too narrow.
image

Especially compared to the header dropdowns:
image

There are also no interaction styling for the Facet items (even a simple pointer cursor).

oleg-semyonov: There seems to be space missing around the `Filters` button on mobile.
image

modules/Search/components/SearchBar.module.scss
```diff
@@ -0,0 +1,77 @@
+.wrapper {
```
oleg-semyonov, line 1: It seems strange that search input is so small on desktop (even smaller than on mobile)

image

oleg-semyonov (changes requested):
This is a great amount of work you've done here!

I have some remarks on the visual part, please check em out. Since Hugo is on vacation, we can ask Al for help if we'll need a designer input on some of my concerns.

modules/Layout/Header/SearchWidget/components/CategoriesList.tsx
fgyimah: Sure will remove this

modules/Search/components/Subtitle.tsx
fgyimah: will add a fixed height for the p element

modules/Search/components/Facet.module.scss
fgyimah: Making the update

oleg-semyonov (approved):
Nice!

PR #278, opened 2022-07-26
[TF-990] added ssr to header to fix the css loading bug

Gichohi-Simon:


oleg-semyonov (changes requested):
This is not a solution I'd expect.

We can't afford to introduce layout shifts to our newsrooms.

Gichohi-Simon:
All right, The other solution I had in mind was to hide the CSS while it's loading, and then show it while it has already loaded and marked it as important in the stylesheet. 
However, I was not so confident about this solution. Should I try this so that you can see what the code looks like?

Gichohi-Simon:
incorrectly named my commit message, but what I meant was that I tried a CSS solution instead of an SSR-based solution. Kindly check the fixing lint error commit for all the changes I made.

oleg-semyonov (commented):
This is still not an optimal solution.
If you'd want to go with CSS approach, the proper solution would be to create an inline `style` tag in the document that would reset the button styles (better yet, inline the `modern-normalize` stylesheets).

The funny thing is that I'm trying to reproduce the issue on the production environment and the styles are loading just fine 🤔 

I think it's better to close this PR and think on the ticket a bit more

Gichohi-Simon:
Okay, let me try and solve it as per your suggestions

PR #280, opened 2022-07-26
[TF-1041] Fix `Quote` and `List` alignments

MohammadxAli:
Same as what I did for the Marcel theme on https://github.com/prezly/theme-nextjs-marcel/pull/143
Applied the same fixes for Bea

### DIFF
--- a/components/RichText/Quote.module.scss
+++ b/components/RichText/Quote.module.scss
@@ -1,3 +1,13 @@
+$content-padding: $spacing-3 + $spacing-4;
+
+@mixin quote-fix($direction) {
+    padding-#{$direction}: $content-padding;
+
+    &::before {
+        #{$direction}: $spacing-4;
+    }
+}
+
 .container {
     @include border-radius-m;
 
@@ -8,17 +18,28 @@
 .content {
     @include paragraph;
 
-    padding: $spacing-4;
     background: $color-base-50;
+    padding: $spacing-4;
+    position: relative;
+
+    &::before {
+        content: "";
+        position: absolute;
+        top: 50%;
+        background-color: $color-base-400;
+        width: 2px;
+        height: calc(100% - ($spacing-4 * 2));
+        transform: translateY(-50%);
+    }
 }
 
 .alignLeft {
     text-align: left;
     justify-content: flex-start;
+    padding-left: $spacing-3;
 
-    .contentInner {
-        padding-left: $spacing-3;
-        border-left: 2px solid $color-base-400;
+    .content {
+        @include quote-fix("left");
     }
 }
 
@@ -26,9 +47,8 @@
     text-align: center;
     justify-content: center;
 
-    .contentInner {
-        padding-left: $spacing-3;
-        border-left: 2px solid $color-base-400;
+    .content {
+        @include quote-fix("left");
     }
 }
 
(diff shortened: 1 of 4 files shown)


components/RichText/Quote.module.scss
```diff
@@ -1,47 +1,43 @@
...
 .alignRight {
     text-align: right;
-    border-right: 2px solid $color-base-400;
-    padding-right: $spacing-3;
+    justify-content: flex-end;
+
+    .contentInner {
+        padding-right: $spacing-3;
+        border-right: 2px solid $color-base-400;
+    }
```
e1himself, line 42: I wonder if this can be solved with `:before`?

I just don't feel comfortable seeing too many wrapper divs for a simple use-case like this :)

components/RichText/Quote.module.scss
```diff
@@ -1,47 +1,61 @@
...
+    padding-#{$direction}: $content-padding;
 
-.left,
-.right {
-    flex-shrink: 0;
-    width: 2px;
-    background-color: $color-base-400;
+    &::before {
+        #{$direction}: $spacing-4;
+    }
```
e1himself, line 8: :)

e1himself (approved):
Tested it. Works well. Good job :+1:

components/RichText/Quote.module.scss
MohammadxAli: @e1himself What do you suggest here? Should I remove the mixin? :) It's really hard to decide if its a good abstraction 😅

e1himself: I just had a smile seeing you fighting even a slight duplication. I'd just leave it as is.

But don't worry, it's good both ways. Just a matter of preference.

MohammadxAli: I really need to read that book from Kent Beck :)

e1himself: > It's really hard to decide if its a good abstraction 

Indeed. That's something you learn over time.

An abstraction has a responsibility. Something it does. Something it _abstracts_ from the developer, something it _encapsulates_ inside itself.

If an abstraction is just a part of a larger code block extracted, it's not an _abstraction_ per se.

Often, it's difficult to find a good name for a "not-so-good" abstraction. For example, `quote-fix` doesn't look like a good name. What is it? A fix? For a quote? What does it do? It's difficult to tell just by looking at it. By inspecting the code I can tell for sure it's not _a fix for a quote_ :)

If I was given a task to refactor it to be a better abstraction, I'd probably do something like this:

```scss
@mixin quote-vertical-line($position: "left") {
      @if ($position == "left") {
        left: $spacing-4;
      } @else {
        right: $spacing-4;
      }
      content: "";
      position: absolute;
      top: 50%;
      background-color: $color-base-400;
      width: 2px;
      height: calc(100% - ($spacing-4 * 2));
      transform: translateY(-50%);
}
```

and then use it as:

```scss
.alignLeft {
   .content {
      padding-left: $content-padding;
      &::before {
         @include quote-vertical-line("left");
      }
   }
}
``` 

Now, `quote-vertical-line` (even though it may still need some love) is almost independent of outer styles and can potentially be even extracted into a global mixin to be used somewhere else. Look how you can read the code now and easily understand what it does without looking into the implementation. Now, that's an abstraction :) 

But generally speaking, I wouldn't do it at all. As the duplication was too simple and too insignificant to be even bothered with it. Also with an abstraction like this (`quote-vertical-line`), the compiler will produce more CSS code. I don't want to accept this trade-off _in this specific case_.

> I really need to read that book from Kent Beck :)

I haven't read it :)

PR #301, opened 2022-08-05
[TF-949] Feature/tf 949 bea greta improvements new contacts

Gichohi-Simon:
1) removed the website link since it was not in the figma design.
2) facebook icon is different from the one on figma, in figma the icon is a square shape and in the code it had 50% border radius: (i checked in the prezly icons, a square one is not available)

### DIFF
--- a/components/ContactCard/ContactCard.module.scss
+++ b/components/ContactCard/ContactCard.module.scss
@@ -36,7 +36,7 @@
     margin: $spacing-2 auto (-$spacing-2);
 }
 
-.contactSection1 {
+.primaryContacts {
     margin: 0 auto $spacing-3;
 
     @include tablet-only {
@@ -67,6 +67,7 @@
 
 .socials {
     display: flex;
+
 }
 
 .icon {
@@ -96,14 +97,6 @@
             margin-bottom: 0;
         }
 
-        .link {
-            margin-right: 0;
-            margin-bottom: $spacing-2;
-
-            &:last-of-type {
-                margin-bottom: 0;
-            }
-        }
     }
 
     /* stylelint-enable max-nesting-depth */
--- a/components/ContactCard/ContactCard.module.scss
+++ b/components/ContactCard/ContactCard.module.scss
@@ -1,19 +1,26 @@
 .avatar {
-    width: 4rem;
-    height: 4rem;
+    width: 3.75rem;
+    height: 3.75rem;
     margin-right: $spacing-3;
     border-radius: 50%;
 }
 
-.content {
-    flex: 1;
-    overflow: hidden;
+.contentTitle {
+    @include tablet-only {
+        display: flex;
+    }
 }
 
 .name {
     @include paragraph;
 
     margin: 0 auto $spacing-half;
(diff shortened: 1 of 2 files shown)


oleg-semyonov:
> removed the website link since it was not in the figma design.

This is not right, it looks like the design just missed this part. We do have website field on the Contacts, so it should be displayed.

components/ContactCard/ContactCard.tsx
```diff
@@ -15,66 +15,64 @@ interface Props {
...
-                            <span className={styles.linkText}>{`@${twitter}`}</span>
-                        </a>
-                    )}
-                    {facebook && (
-                        <a href={`https://facebook.com/${facebook}`} className={styles.link}>
-                            <IconFacebook width={16} height={16} className={styles.icon} />
-                            <span className={styles.linkText}>{facebook}</span>
-                        </a>
-                    )}
+                    <div className={styles.contactSection1}>
```
oleg-semyonov, line 39: Not the best naming IMO. Let's call it `primaryContacts`

components/ContactCard/ContactCard.tsx
```diff
@@ -15,66 +15,64 @@ interface Props {
...
-
     return (
         <div
             id={`contact-${contact.uuid}`}
             className={classNames(styles.container, className, {
                 [styles.compact]: isCompact || device.isMobile,
             })}
         >
-            {renderAvatar({ className: styles.avatar })}
             <div className={styles.content}>
```
oleg-semyonov, line 30: This wrapper is no longer necessary. The classname also doesn't exist anymore, so it does nothing. It's safe to remove it.

components/ContactCard/ContactCard.tsx
```diff
@@ -15,66 +15,64 @@ interface Props {
     isCompact?: boolean;
     renderAvatar: ({ className }: { className: string }) => ReactNode;
 }
-
 function ContactCard({ className, contact, isCompact = false, renderAvatar }: Props) {
     const device = useDevice();
-    const { name, description, company, email, phone, mobile, website } = contact;
```
oleg-semyonov, line 21: Please restore the `website` field. It should be displayed.

components/ContactCard/ContactCard.module.scss
```diff
@@ -45,30 +60,33 @@
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+    font-weight: $font-weight-bold;
+    font-size: $font-size-labels;
+    color: $color-blue-500;
+}
+
+.socials {
+    display: flex;
```
oleg-semyonov, line 69: As I pointed out in Slack, the social contacts alignment issue comes from uneven on the items. In this case it comes from `:last-of-type` rule for the `.link` class. Note that you need to disable that for all cases when the item is displayed inline (e.g. in `socials` and in tablet view for the phone/email section)

oleg-semyonov (changes requested):

oleg-semyonov (approved):
Looks good. Please apply the same changes to Greta.

PR #308, opened 2022-08-08
[MT-5280] Feature - Implement banner on preview newsrooms

e1himself:
Depends on https://github.com/prezly/theme-kit-nextjs/pull/93

Screenshot from 2022-08-08 18-27-55


modules/Layout/Notifications.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Notification } from '@prezly/sdk';
+
+import styles from './Notifications.module.scss';
+import classNames from 'classnames';
+
+interface Props {
+    notifications: Notification[];
+}
+
+export default function Notifications({ notifications }: Props) {
```
e1himself, line 11: @oleg-semyonov 
We'll need the exact same `Notifications` components in all the themes. Do you think it makes sense to share the same implementation via the theme-kit package?

modules/Layout/CookieConsentBar/CookieConsentBar.module.scss
```diff
@@ -1,6 +1,8 @@
+@import "variables";
```
oleg-semyonov, line 1: No need to import anything in SCSS modules, everything needed is injected via the SCSS config.

modules/Layout/Notifications.tsx
oleg-semyonov: Yeah, it makes sense to extract it. Although the theme-kit doesn't enforce any styling (it only compiles JS), so if this components is mostly about displaying the stuff, it should either remain in the theme, or be moved to `themes-ui-components` package.

oleg-semyonov (commented):
Made some remarks, also please note the TSC and ESLint complaints :)

modules/Layout/CookieConsentBar/CookieConsentBar.module.scss
e1himself: Yeah, you're right that it does work. My IDE warns about missing imports though. I'll remove the imports and will just ignore the warnings. Thanks!

Screenshot from 2022-08-10 11-41-56

oleg-semyonov (approved):

PR #322, opened 2022-08-16
[QRF-9] Error with dynamic hrefs on search page

fgyimah:


modules/Search/components/Hit.tsx
```diff
@@ -22,9 +22,14 @@ function HitComponent({ hit }: Props) {
     const { categories } = story;
     const { showDate, showSubtitle } = useThemeSettings();
 
+    const storyLink = {
```
oleg-semyonov, line 25: Please add a comment indicating why this change was needed. That will protect this from future refactorings :)

oleg-semyonov (approved):
Nice! One minor comment to address before merging

modules/Search/components/Hit.tsx
fgyimah: Added the comment!

PR #389, opened 2022-09-28
[QRF-299] Fix - Search results month filter is filtering on the wrong month

OskarGasiorowski:


oleg-semyonov (approved):
Nice! Could you update this code in other themes as well?

OskarGasiorowski:
> Nice! Could you update this code in other themes as well?

@oleg-semyonov should I do this manually or there is a way to do it smarter? :)

oleg-semyonov:
> should I do this manually or there is a way to do it smarter? :)

Unfortunately not, for now it’s all manual changes :/

PR #410, opened 2022-10-11
[DEV-9019] Render new gallery description in themes

fgyimah:


package.json
```diff
@@ -31,9 +31,9 @@
     "@prezly/analytics-nextjs": "1.3.0",
     "@prezly/content-renderer-react-js": "0.24.3",
     "@prezly/icons": "0.5.0",
-    "@prezly/sdk": "7.10.0",
+    "@prezly/sdk": "7.11.0-0",
     "@prezly/story-content-format": "0.58.1",
-    "@prezly/theme-kit-nextjs": "1.17.2",
+    "@prezly/theme-kit-nextjs": "^1.18.0",
```
oleg-semyonov, line 36: Please use `--save-exact` :)

oleg-semyonov (approved):
Looks good! One small note from me, but not blocking (Renovate will pin the deps either way, but nice to keep the pinning in mind when updating deps)

package.json
fgyimah: sure! noted will update before merge

MohammadxAli (approved):

PR #535, opened 2023-01-25
[QRF-856] Boilerplate contact - fixed visibility of a website info

OskarGasiorowski:


modules/Layout/Boilerplate/Boilerplate.tsx
```diff
@@ -58,7 +58,7 @@ function Boilerplate() {
                             )}
                         </div>
                     )}
-                    {hasContactInformation && (
+                    {(hasContactInformation || companyInformation.website) && (
```
OskarGasiorowski, line 61: I was wondering if it's better to update `hasAnyContactInformation` from `theme-kit-nextjs`, but there is a question if Website should be part of the contact, and if changing it wouldn't break something else.

oleg-semyonov: I think it's okay to update the helper in `theme-kit`. This block is pretty much the same across all themes, so should be safe to update.

oleg-semyonov: Also, it would be nice to move the `companyInformation.website` check above together with other checks (e.g. `hasEmail`) to keep things consistent

oleg-semyonov (approved):

PR #591, opened 2023-03-20
[CARE-1215] Get rid of `@prezly/themes-ui-components`

oleg-semyonov:
This PR ports back all of the components from the UI library we extracted some time ago.

"Dumb" UI components were ported into the new `ui` folder:
- Button (removed unused `isActive` and `activeClassName` props)
- FormInput
- LoadingBar
- Modal
- ScrollToTopButton

Other components were ported into `components`:
- CategoryLink (simplified styles to reduce the necessary overrides)
- NotificationsBar
- SocialMeda
- StoryPublicationDate

Additionally, I removed unused color swatches from SCSS colors file, and fixed issues with specificity that arose after the port.

### DIFF
--- a/components/CategoriesList/CategoriesList.module.scss
+++ b/components/CategoriesList/CategoriesList.module.scss
@@ -5,36 +5,20 @@
     font-weight: $font-weight-bold;
 }
 
+.categoryLink {
+    &:hover,
+    &:focus,
+    &:active {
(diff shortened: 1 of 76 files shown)


MohammadxAli (approved):
Nice! I compared it visually and couldn't find any regression.

PR #643, opened 2023-04-21
[CARE-1460] Upgrade to React 18 and Next 13

oleg-semyonov:


.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    extends: ['@prezly', '@prezly/eslint-config/react', '@prezly/eslint-config/nextjs'],
+    extends: ['@prezly/eslint-config/nextjs'],
```
oleg-semyonov, line 2: I updated the configs in v5 to extend each other, so now only a single config for the specific environment is needed.

components/StoryLinks/StoryShareUrl.module.scss
```diff
@@ -27,10 +27,7 @@
     position: absolute;
     transition: opacity 0.4s ease;
     will-change: opacity;
-    left: -9rem;
-    right: 0;
-    top: 0;
-    bottom: 0;
+    inset: 0 0 0 -9rem;
```
oleg-semyonov, line 30: This is a new rule in the standard stylelint config, and it might need getting used to, but it's a bit cleaner with this notation.

fgyimah (approved):

PR #687, opened 2023-05-15
Bump `@prezly/theme-kit-nextjs` to `v5.2.3` and `@prezly/theme-kit-core` to `v5.2.2`

inf1nite-lo0p:


e1himself (approved):

e1himself:
BTW, we have to do the same update to other themes. As they are broken in the same way too.

MohammadxAli:
> BTW, we have to do the same update to other themes. As they are broken in the same way too.

At it :D

MohammadxAli:
Renovate takes care of the other repositories, such as:

- https://github.com/prezly/theme-nextjs-greta/pull/568

PR #694, opened 2023-05-18
[CARE-1673] Switch from `npm` to `pnpm`

MohammadxAli:
Followup on https://github.com/prezly/theme-kit-js/pull/293

[Thanks to @e1himself's awesome script](https://github.com/prezly/theme-nextjs-bea/pull/694/commits/ed3aa9b39d499133a26098a21f96587f3383ec22), we can link a theme to theme-kit the same way we do it for slate in the app.

Imagine having a structure like this:

```
projects/
├─ theme-kit-js/
├─ themes/
│  ├─ theme-nextjs-bea/
```

Then you should symlink the theme-kit and its corresponding packages like this:
```
mkdir linked_modules
cd linked_modules
ln -s ../../../theme-kit-js/
ln -s ../../../theme-kit-js/packages/core/
ln -s ../../../theme-kit-js/packages/nextjs/
```
This is a one-time thing, you don't need to do the above anymore.

Now run the `link-modules` makefile script in the theme's root:
```
make link-modules
```
It will modify `pnpm-workspace.yaml` and `package-lock.yaml` and install packages with `workspace:*` automatically and start the dev server for you.

Once you're done, run:
```
make unlink-modules
```
And it will revert all the actions done in `link-modules` step.


e1himself (commented):
Looks good to me :+1: 

But let's wait for Oleg's review, as the main mastermind behind our themes :)

e1himself:
@oleg-semyonov how do you feel about these changes?

oleg-semyonov:
> @oleg-semyonov how do you feel about these changes?

I feel like it's too much boilerplate just for the ability to link modules 🤔 
Keep in mind that this setup would need to be copied over to other themes, and then maintained.
I'm not sure if the burden of maintaining this setup (which can eventually break the same way as `npm link`) is worth it compared to releasing preview versions of Theme Kit to test big changes.

We might be better off making themes require less library boilerplate altogether (with your work on the Next 13 app router framework), and make theme kit less dependent on React (which will allow us to safely use `npm link`).
Frankly, currently Theme Kit only exports two components, which are for SEO, and would be superseded by `generateMetadata` functions in the app router. So we might be able to drop react dependency entirely 🤔

MohammadxAli:
We came to conclusion to cancel this, as it's not necessary yet. [(more...)](https://linear.app/prezly/issue/CARE-1673/themes-migrating-bea-from-npm-to-pnpm-simplifying-theme-kit-module#comment-5329fbe8)

PR #739, opened 2023-06-07
[DEV-7539] Render stories in the new V4 format

kudlajz:
What happened in this PR:
- updated theme kit to new major version due to usage of SDK v16
  - stories are now fetched in the V4 format due to the changes in theme kit
- implemented `HeaderRenderer` component that renders story Title and Subtitle
- renamed `SlateRenderer` to `ContentRenderer` to better reflect its purpose now that all stories are in Slate format
  - "header" image is now rendered as part of the content


modules/Story/Story.tsx
```diff
@@ -31,47 +30,25 @@ function Story({ story }: Props) {
...
         <Layout>
             <StorySeo story={story} noindex={!isAnalyticsEnabled} />
             <article className={styles.story}>
-                <div
-                    className={classNames(styles.container, {
-                        [styles.withImage]: hasHeaderImage,
-                    })}
-                >
+                <div className={styles.container}>
+                    {isEmbargoStory(story) && <Embargo story={story} />}
```
kudlajz, line 42: Because header image is now rendered via the `ContentRenderer`, it would require extra work to put Embargo in between the header image and story content (as you would have to extract the image element and render it separately).

After a discussion, we've decided to move the Embargo panel to the top of the story, which also increases its visibility (ignore the expanded image, it's going to be contained in the live version):

localhost_3000_i-listened-to-animals-hearts-to-reveal-their-hidden-emotional-worlds (1)

components/ContentRenderer/ContentRenderer.tsx
```diff
@@ -30,13 +30,13 @@ import {
...
 
-import styles from './SlateRenderer.module.scss';
+import styles from './ContentRenderer.module.scss';
 
 interface Props {
     nodes: Node | Node[];
 }
 
-function SlateRenderer({ nodes }: Props) {
+function ContentRenderer({ nodes }: Props) {
```
kudlajz, line 39: I was hesitant to rename the file due to the amount of changes it would introduce in the PR but I think it makes it clearer now that we only have one way of rendering content.

modules/Story/HeaderRenderer/HeaderRenderer.module.scss
```diff
@@ -0,0 +1,12 @@
+.renderer {
+    :global {
+        /* stylelint-disable selector-class-pattern */
```
alpower, line 3: I noticed in other PRs you had added `, max-nesting-depth ` to the `stylelint-disable` - do you need that here as well?

modules/Story/Story.tsx
```diff
@@ -31,47 +30,25 @@ function Story({ story }: Props) {
...
     const hasCategories = categories.length > 0;
+    const nodes = JSON.parse(story.content);
 
     return (
         <Layout>
             <StorySeo story={story} noindex={!isAnalyticsEnabled} />
             <article className={styles.story}>
-                <div
-                    className={classNames(styles.container, {
-                        [styles.withImage]: hasHeaderImage,
```
oleg-semyonov, line 45: I believe this class can be dropped from the stylesheet, since it was there for when the image was above the title/subtitles

modules/Story/HeaderRenderer/Subtitle.module.scss
```diff
@@ -0,0 +1,19 @@
+.subtitle {
```
oleg-semyonov, line 1: I wonder why you didn't style the subtitle with global classes as you did with the title?

modules/Story/HeaderRenderer/HeaderRenderer.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export function HeaderRenderer({ nodes }: Props) {
+    return (
+        <div className={styles.renderer}>
+            <Renderer nodes={nodes} defaultComponents={false} defaultFallback="ignore">
+                <Component match={DocumentNode.isDocumentNode} component={Elements.Document} />
+                <Component match={HeadingNode.isTitleHeadingNode} component={Elements.Heading} />
+                <Component match={HeadingNode.isSubtitleHeadingNode} component={Subtitle} />
+                <Component match={Text.isText} component={Elements.Text} />
+                <Component match={LinkNode.isLinkNode} component={Elements.Passthru} />
```
oleg-semyonov, line 21: This basically means that we're ignoring any links and just render them as text, right?

oleg-semyonov (commented):
Great stuff! I got a couple questions before I can hit that green button :)

oleg-semyonov (approved):
Answered my question myself, so no blocking comments from me 👍🏻

oleg-semyonov (approved):
Answered my question myself, so no blocking comments from me 👍🏻

e1himself (approved):

modules/Story/HeaderRenderer/HeaderRenderer.module.scss
kudlajz: It's not needed here because there are no nested rules.

modules/Story/Story.tsx
kudlajz: Do you mean the `container` or `withImage` class? Because I've removed the `withImage`.

modules/Story/HeaderRenderer/HeaderRenderer.module.scss
kudlajz: > All code/CSS changes make sense to me. Will leave approval to those more familiar with React.

I was more interested in the design review but appreciate your coding skills as well 😛

PR #774, opened 2023-06-22
[DEV-11017] Implement new `Made with Prezly` badge in public themes

fgyimah:


components/MadeWithPrezly/MadeWithPrezly.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useEffect, useState } from 'react';
+
+import { LogoPrezly } from 'icons';
+
+import styles from './MadeWithPrezly.module.scss';
+
+const FOOTER_HEIGHT = 88;
+
+export function MadeWithPrezly() {
+    const [isMadeWithPrezlyVisible, setIsMadeWithPrezlyVisible] = useState(true);
```
oleg-semyonov, line 11: Since this state is encapsulated in the component, you can use shorter names like `isVisible` :)

components/MadeWithPrezly/MadeWithPrezly.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    useEffect(() => {
+        function madeWithPrezlyListener() {
```
oleg-semyonov, line 14: It's more intuitive to simply call this `handleScroll`

components/MadeWithPrezly/MadeWithPrezly.tsx
```diff
@@ -0,0 +1,43 @@
...
+        }
+        if (typeof window !== 'undefined') {
+            window.onscroll = madeWithPrezlyListener;
+        }
+
+        return () => {
+            if (typeof window !== 'undefined') {
+                window.onscroll = null;
+            }
+        };
```
oleg-semyonov, line 28: This is not good, this way of attaching listeners can conflict with browser extensions or other code. You should use `target.addEventListener` to attach events.
Also, you don't need `typeof window` checks in `useEffect`, since it never runs on the server.
Also also, good idea to add `passive: true` to the listener options to indicate that this listener would not interfer with scrolling.

```suggestion
        window.addEventListener('scroll', madeWithPrezlyListener, { passive: true });

        return () => {
            window.removeEventListener('scroll', madeWithPrezlyListener, { passive: true });
        };
```

oleg-semyonov (changes requested):
Got some code-style remarks

components/MadeWithPrezly/MadeWithPrezly.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const [isVisible, setIsVisible] = useState(true);
+
+    useEffect(() => {
+        function scrollListener() {
+            setIsVisible(
+                window.innerHeight + Math.round(window.scrollY) <
+                    document.body.offsetHeight - FOOTER_HEIGHT,
+            );
+        }
+        if (typeof window !== 'undefined') {
```
oleg-semyonov, line 20: As I said, these checks are redundant :)

oleg-semyonov (approved):
👍🏻

pestaa (approved):

PR #872, opened 2023-09-06
[CARE-2137] Date and social icons alignment in themes

fgyimah:
Screenshot 2023-09-06 at 12 43 56 PM
Screenshot 2023-09-06 at 12 43 26 PM

### DIFF
--- a/modules/Story/Story.module.scss
+++ b/modules/Story/Story.module.scss
@@ -19,7 +19,6 @@ $story-container-max-width: 720px;
 .linksAndDateWrapper {
     display: flex;
     flex-direction: column;
-    align-items: center;
 
     &.left {
         align-items: flex-start;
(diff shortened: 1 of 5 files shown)


modules/Story/Story.module.scss
```diff
@@ -13,4 +13,23 @@ $story-container-max-width: 720px;
     @include text-label;
 
     color: $color-base-500;
+    margin: 0;
+}
+
+.linksAndDateWrapper {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
oleg-semyonov, line 22: The default is left alignment, so it's better to define it in the styles. Or you could simply omit this line, which will do pretty much the same.

oleg-semyonov (approved):
One minor suggestion from me

modules/Story/Story.module.scss
fgyimah: Updated https://github.com/prezly/theme-nextjs-bea/pull/872/commits/d1f00798de93b1941d68af8eee97d0bd26bc54d7

PR #888, opened 2023-09-20
[CARE-3129] Preserve line breaks in address

kudlajz:
Screenshot 2023-09-20 at 11 14 17

### DIFF
--- a/modules/Layout/Boilerplate/Boilerplate.module.scss
+++ b/modules/Layout/Boilerplate/Boilerplate.module.scss
@@ -55,6 +55,7 @@
 
     display: flex;
     align-items: flex-start;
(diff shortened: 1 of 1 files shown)


modules/Layout/Boilerplate/Boilerplate.module.scss
```diff
@@ -55,6 +55,7 @@
 
     display: flex;
     align-items: flex-start;
+    white-space: pre;
```
kudlajz, line 58: I'm still not sure what happened here. The address string always had line breaks in it as far as I'm aware, so I wonder if this even worked before? And if it did, how would that be possible? 🤔

PR #912, opened 2023-10-13
[DEV-11785] Next.js 13 - Experiment with custom routing and data-fetching PoC

e1himself:
- Use the new Next.js 13 app router
- Apply Next.js 13 codemods to migrate existing markup to the updated behavior
- Use ServerComponents for page rendering
- Use the async rendering data-fetching strategy -- by means of the Content Delivery optimized wrapper around Prezly API SDK (based on PrezlyAPI from the `theme-kit` package)

Things to look at:

- `/theme/_router.ts` file for wiring the page modules to routes
- `/app/[[...path]]` folder where routing logic is wired to the Next.js app router
- `/theme` folder for the pages markup 
- `/theme-kit` for prototype code that will become a part of the shared toolkit package


app/[[...path]]/page.tsx
```diff
@@ -0,0 +1,12 @@
...
+import router from '@/theme/_router';
+
+export default async function Page(props: { params: { path?: string[] } }) {
+    const { path = [] } = props.params;
+
+    const { match, Page: PageComponent } = await router(path);
+
+    // @ts-expect-error Server Component
+    return createElement(PageComponent, match);
+}
```
Comment (e1himself) [inline app/[[...path]]/page.tsx:12]: So I've used this catch-all Next.js route to workaround the limitations of ServerComponents and reduce each theme boilerplate connecting the newsroom markup to the Next.js app instance.

The limitations with Next.js 13 server components I had to workaround are:

- layout components cannot access path parts below its level
   - root layout will never know which locale it's being rendered for; the only way to get the first `/[locale]` subpath value is to introduce a second level of layout (`app/[locale]/layout.tsx`)
   - there cannot be multiple dynamic subpath routes on the hierarchy level: it's not possible to have `/[locale]` and `/[slug]` subpaths with an additional matching logic -- deciding whether it is a localized homepage or a story page.


The additional routing layer provides some benefits:

- abstracts the generic catch-all route complexity by providing a human-friendly API in binding page modules and layouts to route params
- provides a way to additionally check the matched params against the API data. E.g.:
   - `/[locale]` route will match only if there is a language enabled for this locale in the newsroom
   - `/[slug]` route will match only if there is a story with this slug

- I wanted to avoid having `/[locale-or-slug]` Next.js as the only route with dynamic rendering (as it'll make things harder to understand) and just went full on the additional dynamic router layer, wiring theme-domain pages to the routes (see `theme/_router.ts`)

theme-kit/api.ts
```diff
@@ -0,0 +1,22 @@
...
+    const client = createPrezlyClient({
+        accessToken: PREZLY_ACCESS_TOKEN,
+        baseUrl: PREZLY_API_BASEURL,
+    });
+
+    return {
+        api: client,
+        contentDelivery: createContentDeliveryClient(client, PREZLY_NEWSROOM_UUID),
+    };
+}
```
e1himself, line 22: This is a helper function that can be used anywhere during the ServerComponent rendering phase, it relies on the `env()` helper to acquire the environment variables

theme-kit/env.ts
```diff
@@ -0,0 +1,80 @@
...
+    })
+    .catchall(z.string());
+
+type ExpectedEnv = z.infer<typeof Schema>;
+
+export function env(): ExpectedEnv {
+    assertServerEnv('env');
+
+    return validateEnv(getEnvVariables(process.env, headers()));
+}
```
e1himself, line 27: This helper inspects `process.env` and `headers()` for relevant values and validates it against the `zod` schema, making sure we don't have any misconfiguration and that the returned object always complies with the expected interface

theme-kit/lib/createContentDeliveryClient.ts
```diff
@@ -0,0 +1,203 @@
...
+
+const DEFAULT_PAGE_SIZE = 20;
+
+interface Params {
+    formats?: Story.FormatVersion[];
+    pinning?: boolean;
+    pageSize?: number;
+}
+
+export function createContentDeliveryClient(
```
e1himself, line 13: This is a simplified reimplementation of `PrezlyAPI` from the `theme-kit` package. The goal is move it to the `theme-kit` repo and replace `PrezlyAPI` with simpler more compact APIs, so it feels natural next to Next.js helpers (naming-wise too)

theme/gallery-album.tsx
```diff
@@ -0,0 +1,29 @@
...
+    uuid: string;
+}
+
+export async function match({ uuid }: Match) {
+    const { contentDelivery } = api();
+
+    const gallery = await contentDelivery.gallery(uuid);
+
+    return gallery && { gallery };
+}
```
e1himself, line 16: Theme page modules can export a `match` function to additionally process the router match params to:

- validate it against the real API data
- enrich the page props with real data objects (to not request them again)

This is necessary to make multiple same-level dynamic routes work without collisions: primarily for `/[locale]` and `/[slug]` routes.

theme/story.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const { contentDelivery } = api();
+
+    const story = await contentDelivery.story(params);
+
+    return story && { story };
+}
+
+export function resolveLocale({ story }: Props) {
+    return story.culture.code;
+}
```
e1himself, line 18: Additionally, a page module may export `resolveLocale()` function to infer the current website locale from a non-`locale` route parameter. Primarily for stories `/[slug]` pages.

theme/story.tsx
```diff
@@ -0,0 +1,50 @@
...
+}) {
+    return (
+        <html lang={props.locale}>
+            <body>
+                <h1>Story layout</h1>
+                {props.children}
+            </body>
+        </html>
+    );
+}
```
e1himself, line 50: Page modules may override Layout by exporting a `Layout` server component. I'm not sure yet if this will be necessary, though.

theme-kit/lib/locales.js
```diff
@@ -0,0 +1,112 @@
+/**
+ * This list is pulled from the main Prezly application. The underscores are replaced with dashes.
+ */
+module.exports = [
```
e1himself, line 4: This is not used yet. Probably, I will remove it, as the router does not need a list of hard-coded locales anymore.

utils/assertServerEnv.ts
```diff
@@ -0,0 +1,5 @@
+export function assertServerEnv(functionName: string) {
```
oleg-semyonov, line 1: FYI there's also `server-only` package that you can import into a module to mark it as such.

theme/_router.ts
```diff
@@ -0,0 +1,19 @@
...
+        route('/', () => import('./homepage')),
+        route('/[locale]', () => import('./homepage')),
+        route('/[slug]', () => import('./story')),
+        route('/s/[uuid]', () => import('./story')),
+        route('/media', () => import('./gallery')),
+        route('/[locale]/media', () => import('./gallery')),
+        route('/media/album/[uuid]', () => import('./gallery-album')),
+        route('/[locale]/media/album/[uuid]', () => import('./gallery-album')),
+        route('/search', () => import('./search')),
+        route('/[locale]/search', () => import('./search')),
```
oleg-semyonov, line 17: It feels a bit weird to have two route definitions.
I know that Next doesn't support a required path parameter after an optional one, but can we do this for static routes like this one?
```suggestion
        route('/[...locale]/search', () => import('./search')),
```

theme-kit/lib/createContentDeliveryClient.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+        },
+
+        stories(params: {
+            search?: string;
+            category?: Pick<Category, 'id'>;
+            locale?: Pick<Culture, 'code'>;
+            page?: number;
+            pageSize?: number;
+            withFeaturedItems?: number;
```
oleg-semyonov, line 91: This parameter seems to be unused

theme-kit/lib/createContentDeliveryClient.ts
```diff
@@ -0,0 +1,203 @@
...
+export function createContentDeliveryClient(
+    prezly: PrezlyClient,
+    newsroomUuid: Newsroom['uuid'],
+    {
+        formats = [Story.FormatVersion.SLATEJS_V4],
+        pinning = false,
+        pageSize: defaultPageSize = DEFAULT_PAGE_SIZE,
+    }: Params = {},
+) {
+    const contentDeliveryClient = {
```
oleg-semyonov, line 22: Wouldn't it be cleaner to use a class instead? IMO it's a bit harder to read in this form (especially when referencing the params)

app/[[...path]]/layout.tsx
```diff
@@ -0,0 +1,20 @@
...
+import router from '@/theme/_router';
+
+export default async function Layout(props: { children: ReactNode; params: { path?: string[] } }) {
+    const { params, children } = props;
+    const { path = [] } = params;
+
+    const { match, Layout: PageLayout } = await router(path);
+
+    if (PageLayout) {
+        // @ts-expect-error Server Component
```
Comment (e1himself) [inline app/[[...path]]/layout.tsx:14]: This is a known issue of Typescript JSX declarations not being (yet) compatible with async ServerComponents model.

The current workaround is to silence these errors with a `@ts` directive:
https://github.com/vercel/next.js/issues/42292#issuecomment-1298483608

Comment (e1himself) [reply app/[[...path]]/layout.tsx:14]: Oh, actually, it seems the newest TS package may be fine with it. I'll try it later.

theme-kit/router.ts
```diff
@@ -0,0 +1,147 @@
...
+                        ...pageMatched,
+                        locale: await resolveLocale(pageMatched),
+                    },
+                    Layout,
+                    Page,
+                };
+            }
+
+            return {
+                match: { ...pageMatched, locale: matched.locale ?? (await getDefaultLocale()) },
```
oleg-semyonov, line 55: This duplicates the logic on line 31

theme-kit/router.ts
```diff
@@ -0,0 +1,147 @@
...
+              Page: ServerComponentType<Props>;
+          }
+        | undefined
+    >;
+};
+
+export async function match(path: string | string[], routes: Route<any, any>[]) {
+    assertServerEnv('match');
+
+    const { contentDelivery } = api();
```
oleg-semyonov, line 81: I found this shorthand naming a bit weird. Maybe I'm too used to `getX` naming for such methods 🤔

theme-kit/router.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    async function getDefaultLocale() {
+        return (await contentDelivery.language())!.locale.code;
```
oleg-semyonov, line 84: I think this should be typed better (e.g. when `localeCode` is not provided, the types should indicate that a value would always be returned).

theme/_router.ts
```diff
@@ -0,0 +1,19 @@
...
+    ]);
```
e1himself, line 18: This may become abstracted one level further: by hiding the actual routes behind the default setup (which is unlikely to change from theme to theme).

theme-kit/router.ts
```diff
@@ -0,0 +1,147 @@
...
+                return undefined;
+            }
+
+            if (matched.locale && !(await isEnabledLocale(matched.locale))) {
+                return undefined;
+            }
+
+            if (!matched.locale) {
+                matched.locale = await getDefaultLocale();
+            }
```
e1himself, line 33: Locale code matching and defaults is baked into the router, so it doesn't have to be implemented for every localized sub-path.

theme-kit/router.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    async function isEnabledLocale(locale: string) {
+        return contentDelivery.language(locale).then(Boolean);
+    }
+
+    const stringPath = typeof path === 'string' ? path : `/${path.join('/')}`;
+    const matches = await Promise.all(
+        routes.map((r) => r.match(stringPath, getDefaultLocale, isEnabledLocale)),
+    );
```
e1himself, line 94: Route matching is done with parallel async execution to evaluate the matched route URLs against the API data without waiting for each response sequentially.

theme/gallery-album.tsx
oleg-semyonov: This particular case doesn't really utilize the server components paradigm, where the data is fetched at the component level. In this current form it looks more like the old `getServerSideProps` 🤔

theme/homepage.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { api } from '@/theme-kit';
+
+interface Match {
+    locale: Culture['code'];
+}
+
+export default async function Page({ locale }: Match) {
+    const { contentDelivery } = api();
+
+    const language = (await contentDelivery.language(locale))!;
```
oleg-semyonov, line 12: Not a big fan of this non-null assertion.
Can we make a separate endpoint for validating the locale (where it's not yet clear if that's an enabled locale), and another one for scenarios we're confident that locale is enabled (like here)?

theme/story.tsx
```diff
@@ -0,0 +1,50 @@
...
+import type { ReactNode } from 'react';
+
+import { api } from '@/theme-kit';
+
+type Match = { slug: string } | { uuid: string };
+
+export async function match(params: Match) {
+    const { contentDelivery } = api();
+
+    const story = await contentDelivery.story(params);
```
oleg-semyonov, line 11: Continuing my comment in `gallery-album.tsx`, I don't get why we can't fetch the story in the Page component and call Next's `notFound()` method when the story is not found?

This would make the props logic much simpler and cleaner.

theme/story.tsx
oleg-semyonov: In all of our current themes, the layout is the same for all pages (except for homepage, which has featured contacts, but that would be better moved to the page itself). So I don't think we'll ever use it.

app/[[...path]]/layout.tsx
```diff
@@ -0,0 +1,20 @@
+import type { ReactNode } from 'react';
+import { createElement } from 'react';
+
+import { Layout as RootLayout } from '@/theme/_layout';
+import router from '@/theme/_router';
+
+export default async function Layout(props: { children: ReactNode; params: { path?: string[] } }) {
```
Comment (e1himself) [inline app/[[...path]]/layout.tsx:7]: I believe this can become a static layout (without a way to override it on the page module level). This way the out-of-the-box Next.js optimizations should still work for layouts.

app/[[...path]]/layout.tsx
Comment (oleg-semyonov) [reply app/[[...path]]/layout.tsx:14]: Yeah, I fixed this error in the website after upgrading to latest TS

theme-kit/router.ts
oleg-semyonov: I wonder if we could simplify this whole thing by using `middleware`.
The only real conflict we have is `/[locale]` vs `/[slug]`.

What if we define a `/story/[slug]` route, then have a Next middleware function process all requests that match `/[locale|slug]` and based on the checks would rewrite to `/story/[slug]`?

theme/gallery-album.tsx
e1himself: > In this current form it looks more like the old getServerSideProps 🤔

You're right that it's similar. But it's also a way to reduce duplication.

An advantage is that this only fetches the route root object (the `gallery` in this case), allowing the underlying components to fetch more data as they need. For example, the `newsroom` object will never be passed down the tree and can be requested via `await contentDelivery.newsroom()` anywhere. Same for `categories`, language boilerplate settings, and so on.

theme/_router.ts
e1himself: Yeah, this is something I definitely wanted to improve here. Because this is actually custom syntax that only resembles Next.js routing conventions, this can be modified to allow optional path variables. Like so:

```
/[[locale]]/search
```

... and the router will just understand it. 

I didn't address it here, as it wasn't required for the minimal PoC I've been aiming for.

app/[[...path]]/layout.tsx
Comment (oleg-semyonov) [reply app/[[...path]]/layout.tsx:7]: According to Next docs, the root `layout` component is required anyway. However, it is only intended to provide the boilerplate HTML, since that will not be re-rendered from route to route.

See https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#root-layout-required

For layouts that should be re-rendered on each request, there are [Template](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#templates) (although not sure if the support RSC)

theme-kit/router.ts
e1himself: > Maybe I'm too used to `getX` naming for such methods 🤔

This is a matter of preference. The academic Java-way of naming everything as `getX` is one approach. Next.js, for example, is preferring shorter names: `headers()` instead of `getRequestHeaders()`, same for `cookies()`. 

I wanted it to be compact and feel native to a Next.js application. And after you adjust the expectations, it looks good, if you ask me. :)

theme/homepage.tsx
e1himself: I agree. I just didn't spend time on typescript gymnastics, leaving it for later. I'll definitely look for ways to avoid these.

theme/story.tsx
e1himself: > I don't get why we can't fetch the story in the Page component and call Next's notFound() method when the story is not found?

Yeah, you can do that in the `gallery-album` module. As there are guaranteed to be no other candidates for that URL.

But it is different here, for the `story` module -- the route is competing with the same-hierarchy-level `/[locale]` route, so we want to cut it off on the earlier stage: before the rendering happens.

Potentially, this can be changed here too, as the router [is locale-aware now](https://github.com/prezly/theme-nextjs-bea/pull/912#discussion_r1358216549) (I've added that logic on rather late stages), so story fetching can be moved to the rendering phase without having any unwanted consequences.

However, as I have already said in [my other comment](https://github.com/prezly/theme-nextjs-bea/pull/912#discussion_r1358233949): there is still a big difference compared to `getServerProps` approach -- as this is only fetching the direct route-bound data objects, allowing the components to fetch the rest using direct API interactions. 

So it's still much simpler. 

Anway, I'll think about this more. Thank you for this remark.

theme-kit/router.ts
e1himself: > and based on the checks would rewrite to `/story/[slug]`?

I don't like this particular aspect of your suggestion. I'd not want to do internal rewriting magic to keep things directly mapped between the public URL and internal structure.

But it's a great point that having parts of this unloaded to a middleware may help. I'll explore it. 

Thanks!

theme-kit/lib/createContentDeliveryClient.ts
e1himself: IDK, for some reason I don't like classes in JS. It feels to me a bit alien to the ecosystem :thinking: 

One particular advantage with this approach, is that there is no need to rely on `this` anywhere in this code. So you can destructure the object and use parts of it wherever you want.

Disregarding the cleanness (which is subjective), are there any performance differences between using both approaches that you are aware of?

oleg-semyonov: None that I'm aware of, this is purely subjective comment from me :)

theme-kit/router.ts
e1himself: Hmm.. :thinking:  Not really. As `pageMatched` is a different object compared to `matched`. 

But I see why you're confused, it's (too) tricky. I'll think on how it can be restructured. :+1:

app/[[...path]]/page.tsx
Comment (e1himself) [reply app/[[...path]]/page.tsx:12]: > and based on the checks would rewrite to `/story/[slug]`?

I don't like this particular aspect of your suggestion. I'd not want to do internal rewriting magic, and instead better keep things directly mapped between the public URL and internal structure.

Also, I think it's important to keep in mind that requirements change over time and may require more conflicting routes later (or even now, for a custom theme) and with this approach, there is some spare flexibility remaining to withstand extensions in the future. I'd not want to tailor-make the routing logic so it can only work for the current fixed set of requirements.

But anyway, it's a great point that having parts of this unloaded to a middleware may help. I'll explore it. Thanks!

Comment (e1himself) [reply app/[[...path]]/page.tsx:12]: > Also, I think it's important to keep in mind that requirements change over time and may require more conflicting routes later

One particular use-case that we've been talking about internally a lot: evergreen content pages (i.e., unlike time-based stories feed, these could be published and displayed as top-level website pages). This feature will most likely require the introduction of another sale-level dynamic route.

theme/story.tsx
e1himself: 57223759e1bbce2cfc9741976c1b05cbfbf9e9d8

app/[[...path]]/layout.tsx
Comment (e1himself) [reply app/[[...path]]/layout.tsx]: fb051caebee910c2e9f70fb1f1530ef857cd304d

app/[[...path]]/layout.tsx
Comment (e1himself) [reply app/[[...path]]/layout.tsx]: FWIW, this _is_ the root layout, as the root catch-all route is viewed by Next.js as the root location.

utils/assertServerEnv.ts
e1himself: 0fa9f52c8344f91f221d5cc86cfa2e1cf3c808b2

theme-kit/lib/createContentDeliveryClient.ts
e1himself: 06f784a86e0d3104cf6c38e422ccee71be73b624

theme-kit/router.ts
e1himself: dc6e0c79c44eae36933d9b3ad9f6657d1e985cca

theme/homepage.tsx
e1himself: dc6e0c79c44eae36933d9b3ad9f6657d1e985cca

theme/gallery-album.tsx
oleg-semyonov: Still, apart from the `/[locale]` - `/[slug]` conflict, I don't see any benefit of this approach over simply fetching the data in the component and calling `notFound` if the gallery is not found.

Also, with this approach, for every request to any route, this data-fetching code will run for every defined route, which is totally unnecessary and introduces extra load on the API without real need for it.

I get that it makes sense for routes that can conflict and should rely on API to resolve, but for any other routes that have a clear static part (e.g. `/s/[uuid]`, `/media/album/[uuid]`, `/search`), there is no point in calling the data-fetching logic until the route is actually matched and rendered..

e1himself: > but for any other routes that have a clear static part (e.g. /s/[uuid], /media/album/[uuid], /search), there is no point in calling the data-fetching logic until the route is actually matched and rendered.

But it will not. The module-level `match()` function is only called if the static route pattern has already matched, and skipped if it does not. So, in fact, this will only make a different for the conflicting routes.

oleg-semyonov: Okay, that it clears it up a bit. 
But still, it's a bit confusing to me, that there's a function called `match` whose primary function is to help the router match the routes that could conflict, but then it somehow also provides data for the page component.

If you're worried about the same request being called twice, Next.js caching should take care of that (since the same function would be called with the same arguments). I agree that `match` is useful, but I'd leave it to only serve as the helper for the router, not a data-fetching method.

e1himself: > If you're worried about the same request being called twice, Next.js caching should take care of that (since the same function would be called with the same arguments)

Yeah, I know. That's why I said that [newsroom object will never be passed down](https://github.com/prezly/theme-nextjs-bea/pull/912#discussion_r1358233949) and can be fetched wherever it is needed.

> but I'd leave it to only serve as the helper for the router, not a data-fetching method.

If we don't allow reusing the data-fetching result of the `match()` function, then you'll need to fetch it again and check if it's not null again:

```
export function match({ slug }) {
  const story = api.story({ slug });
  return Boolean(story);
}

export default function Page({ slug }) {
  const story = api.story({ slug });

  if (!story) {
     notFound();
  }

  return <Markup story={story} />
}
```

And you're right that moving it to the render function _in other modules_ is possible (and may be preferrable). But that's unimportant. I see it as a nice side-effect of route conflict resolution logic that allows me do skip throwing `notFound()` manually in the render component. But this is just an implementation detail.

PR #929, opened 2023-10-25
[DEV11828] Next.js 13 - Move locale detection logic to middleware

e1himself:
This is a second iteration of #912 

- The custom routing logic has been moved from the view layer (`Page` & `Layout`) to the middleware layer
  - This allows writing additional logic of locale detection (`Accept-Language` header + cookies, as `next-intl` does)
  - Out-of-the-box `next-intl` middleware does more than we need and is too opinionated to  be tailored to our use-case, so I'm not using it
 
- Routing is done by rewriting URLs internally to route requests to the right `app` page. So all Next.js file conventions and optimizations work out-of-the-box. 
- The `app` page files hierarchy is now close to what production URLs look like. With the addition of always having the full locale code present as the `[locale]` prefix for all pages (makes it possible to provide i18n based on request path params)


theme-kit/env.ts
```diff
@@ -20,10 +25,32 @@ const Schema = z
...
+
+    return environment;
 }
 
+export function env() {
+    const cached = getEnvironment();
+    if (cached) {
+        console.info('Reusing cached HTTP env object.');
+        return cached;
+    }
```
e1himself, line 44: HTTP environment is now cached per request-render cycle and is no longer needed to be reconstructed every time it is required.

Thanks to [`React.cache()`](https://react.dev/reference/react/cache) (used internally by `server-only-context`

app/router.ts
```diff
@@ -0,0 +1,29 @@
+import 'server-only';
+
+import { api, createRouter, route } from '@/theme-kit';
+
+export default function createAppRouter() {
```
e1himself, line 5: This router, along with the middleware, will become a part of the shared theme-kit logic that we won't be needing to reimplement in every theme.

next.config.js
```diff
@@ -11,8 +11,12 @@ const globalSassImports = `\
     @import "styles/mixins";
 `;
 
+function identity(value) {
+    return value;
+}
+
 const moduleExports = withBundleAnalyzer(
-    withPrezlyConfig({
+    identity({
```
e1himself, line 19: I've temporarily disabled this config to make sure it does not interfere with the new logic.

theme-kit/middleware.ts
```diff
@@ -0,0 +1,96 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+
+import { api } from './api';
+import { locale } from './locale';
+import { createRouter, route } from './router';
+
+export async function middleware(request: NextRequest) {
```
e1himself, line 9: The middleware, along with the router configuration, will become a part of the shared theme-kit logic that we won't need to reimplement in every theme.

theme-kit/middleware.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function createAppRouter() {
+    const { contentDelivery } = api();
+
+    return createRouter([
+        route('/(:localeSlug)', '/:locale'),
+        route('(/:localeSlug)/category/:slug', '/:locale/category/:slug'),
+        route('(/:localeSlug)/media', '/:locale/media'),
+        route('(/:localeSlug)/media/album/:uuid', '/:locale/media/album/:uuid'),
+        route('(/:localeSlug)/search', '/:locale/search'),
```
e1himself, line 72: These are using the standard [`URLPattern`](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/URLPattern)  Web API interface, provided by the [`url-pattern`](https://www.npmjs.com/package/url-pattern) package.

package.json
```diff
@@ -46,11 +47,12 @@
     "modern-normalize": "1.1.0",
     "next": "13.5.4",
     "parse-data-url": "6.0.0",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
+    "react": "18.3.0-canary-09fbee89d-20231013",
+    "react-dom": "18.3.0-canary-09fbee89d-20231013",
```
e1himself, line 51: I had to switch to React `@canary` version in order to use [`React.cache()`](https://react.dev/reference/react/cache). This provides the closest thing we can get to a server-components context. This is also how `next-intl`'s [`unstable_setRequestLocale`](https://next-intl-docs.vercel.app/docs/getting-started/app-router-server-components#add-unstable_setrequestlocale-to-all-layouts-and-pages) works.

Not ideal. Potentially, this can be reverted by always recalculating the env object on demand, as previously. But I wanted to hear your opinion on it first.

oleg-semyonov (commented):
I like this new iteration, looks much cleaner 👍🏻 

Would be nice to have the deployment working so we can see it in action

theme-kit/middleware.ts
```diff
@@ -0,0 +1,96 @@
...
+             * - _next/image (image optimization files)
+             * - favicon.ico (favicon file)
+             */
+            '/',
+            '/((?!api|_next/static|_next/image|favicon.ico).*)',
+        ],
+    };
+}
+
+function createAppRouter() {
```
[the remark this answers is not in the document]
e1himself: Oh, this is a leftover. It can be safely removed. Thanks!

97bd01f675e187294d22a0cc5363e9ab6618bc00

theme-kit/env.ts
e1himself: Yeah, I wasn't completely sure if it's worth it. I guess, we'll only know for sure after the real performance tests.

One of the reasons I thought it is necessary was to pass the currently active locale down the tree (as it was in the context; see `unstable_setRequestLocale()` from `next-intl`). But I've found a better workaround for it by using the request header.

So, I guess, I can revert back to the stable React build and reconsider it again after we have it running in a close-to-prod environment.

Thanks!

e1himself: 3eff05e8c7486bccd8269776e3366461ed41108a

package.json
e1himself: Reverted back to the stable version.

See:

- https://github.com/prezly/theme-nextjs-bea/pull/929#discussion_r1371523689
- 43c56696b3df051b1232fdf0fe206f5caf1c3bea
- 3eff05e8c7486bccd8269776e3366461ed41108a

app/[locale]/[slug]/page.tsx
```diff
@@ -0,0 +1,20 @@
+import { notFound } from 'next/navigation';
+
+import { api } from '@/theme-kit';
+
+interface Params {
```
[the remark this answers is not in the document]
Comment (e1himself) [reply app/[locale]/[slug]/page.tsx]: f65c4d153217434e8ea378c2ca4e21e76accdcf7

theme-kit/env.ts
```diff
@@ -20,10 +25,32 @@ const Schema = z
...
+
+    console.warn(
+        'Rebuilding process+HTTP env, as there is no data in the cache available.' +
+            'Either because bootHttpEnv() func was not called in the root Layout,' +
+            'or because the env() function is called outside of React tree.',
+    );
```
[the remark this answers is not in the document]
e1himself: This is not relevant anymore, after reverting the code back to the stable React build.

3eff05e8c7486bccd8269776e3366461ed41108a

PR #931, opened 2023-10-25
[DEV-11815] Next.js 13 - Convert theme to Next.js 13

e1himself:


modules/Story/index.ts
```diff
@@ -1 +1,2 @@
-export { default } from './Story';
+// default export works better with Next.js `dynamic()` component import
+export { Story } from './Story';
```
MohammadxAli, line 2: This isn't a _default_ export though. The component is refactored to use named export instead, so maybe this comment is outdated?

modules/Search/components/Results.tsx
```diff
@@ -1,58 +1,60 @@
...
-    const { formatMessage } = useIntl();
-    const { searching: isSearching } = useAlgoliaState();
-
-    return (
-        <div className={containerStyles.container}>
-            <div className={classNames(listStyles.storiesContainer, styles.list)}>
-                {!hits.length && (
-                    <p className={styles.fallbackText}>
+import containerStyles from '@/modules/InfiniteStories/InfiniteStories.module.scss'; // FIXME: Pass this from outside
+import listStyles from '@/modules/InfiniteStories/StoriesList.module.scss'; // FIXME: Find a way to pass this from ouside
```
MohammadxAli, line 17: 👀

Leaving a comment just so we don't forget about it.

modules/Search/components/Hit.tsx
```diff
@@ -40,35 +60,32 @@ function HitComponent({ hit }: Props) {
             <div className={cardStyles.content}>
                 {categories.length > 0 && (
```
MohammadxAli, line 61: Shouldn't this check for `displayedCategories.length` instead? As we're passing `displayedCategories` below.

```suggestion
                {displayedCategories.length > 0 && (
```

MohammadxAli (commented):
This morning I planned to check this PR so that I can learn something :)

Reviewed 1/3 of the files and I have to say: "This is just magnificent work Ivan! 👍💯"

IMO, the code is a lot easier to reason about even though it might seem more complex at a glance.

modules/Search/components/Hit.tsx
e1himself: Good catch! Thanks! :+1:

modules/Search/components/Results.tsx
e1himself: Yeah, but this is not for now. We'll address it while switching to the Design System UI components we've prepared.

modules/Story/index.ts
e1himself: Good point. Will remove it :+1: 

Thanks!

e1himself: e06f8a96490901da6ae3cca3220cc0cbad04c5d2

modules/Search/components/Hit.tsx
e1himself: e06f8a96490901da6ae3cca3220cc0cbad04c5d2

PR #932, opened 2023-10-25
[DEV-11815] Next.js 13 - Server-side i18n utils

e1himself:
- Replace the deprecated `@prezly/themes-intl-messages` packages with new `@prezly/theme-kit-intl`
- Implement own server-side `FormattedMessage` component to replace its counterpart from `react-intl`
- Replace `FormattedMessage` usages with the local implementation
 
  _**Note**:_ _The `FormattedMessage` implementation in this PR is server-side only, while most of the UI components using it here are still client-side. It is supposed to do the conversion in subsequent pull requests._


theme-kit/intl.tsx
```diff
@@ -0,0 +1,72 @@
...
+type Dictionary = Record<string, MessageFormat>;
+type MessageFormat = { type: 0 | 1; value: string }[];
+
+async function importDictionary(code: Locale.Code): Promise<Dictionary> {
+    const fileCode = getSupportedLocaleIsoCode(code);
+
+    return import(`@prezly/theme-kit-intl/messages/${fileCode}.json`);
+}
+
+function formatIntlMessage(
```
oleg-semyonov, line 50: `react-intl` also has `formatMessage` function returned from `useIntl` hook, that is used for places where JSX is not supported (e.g. `title` property on a link). 
Would it be possible to also implement that here? Then we'll likely be able to drop the `react-intl` dependency completely.

oleg-semyonov (commented):
Looks good! I have one proposition :)

theme-kit/intl.tsx
e1himself: > Then we'll likely be able to drop the react-intl dependency completely.

Yeah, I have been thinking about this too :) 

Btw, the server-components version of this function is 10 lines above your comment :) 

https://github.com/prezly/theme-nextjs-bea/blob/71e397636bb1e869d1e5813e0d7fa4c11d25464e/theme-kit/intl.tsx#L31-L39

oleg-semyonov: > Btw, the server-components version of this function is 10 lines above your comment :)

Yes, but from what I saw in the code, it still renders JSX, right? We also need a version that returns a plain string for the cases I described above

e1himself: > We also need a version that returns a plain string for the cases I described above

Ah, that's a good point. I thought that `formatMessage()` and `FormattedMessage` are doing the same thing. But now I see they don't. Thank! : )

e1himself: 910d2243215bdc558f05b0d070bdb8c79ac759b3

PR #1032, opened 2023-12-14
[DEV-12121] Feature - Use the new slugs history api in the selected newsroom

e1himself:
This is based on the code from https://github.com/prezly/theme-kit-js/pull/511

I will release the next major version if you're happy with how this looks/works.


oleg-semyonov (approved):
I assume the rest of the changes are kept hidden behind the page-props helpers, which didn't change their API signature, right?

e1himself:
> I assume the rest of the under-the-hood Theme Kit changes are kept hidden behind the page-props helpers, which didn't change their API signature, right?

Yes, that is correct.

PR #1039, opened 2023-12-25
Experiment: use Biome for code style

riffbyte:
Our code-style setup has been overly complex for its own good, and trying to make so much tools and plugins work together is a big challenge, so I've been wondering if we could use a simpler tool.

This PR makes use of [Biome](https://biomejs.dev/) - a new all-in-one tool for linting and formatting. While it's pretty new and isn't 100% matching our current setup in terms of rules and their functionality, it does cover most of the requirements, and it is also WAAAY faster (linting only takes 20-30ms with Biome, while ESLint takes several seconds to run).

A couple drawbacks I found while experimenting:
- No option to ignore a rule for the whole file (only through the config file), requiring adding suppress comments to every line individually (a painful case with naming convention rule)
- Some rules don't have enough options to configure them to our use-case (this is by design, so far naming convention is the biggest offender)
- While the package is pretty new, false-positives happen (encountered one for exhaustive deps)
- Auto-fixing doesn't always work as intended (might be something wrong with the VS Code extension though)

Would appreciate your opinions on this one! Perhaps this would be a good time to reconsider some of the rules that we're using, maybe we don't even need that much?

### DIFF
--- a/@types/global.d.ts
+++ b/@types/global.d.ts
@@ -1,8 +1,10 @@
 import type { PrezlyEnv } from '@prezly/theme-kit-nextjs';
 
 declare global {
+    // biome-ignore lint/style/useNamingConvention: <explanation>
     export namespace NodeJS {
         export interface ProcessEnv extends PrezlyEnv {
+            // biome-ignore lint/style/useNamingConvention: <explanation>
             NEXT_PUBLIC_HCAPTCHA_SITEKEY: string;
         }
     }
--- a/@types/images.d.ts
+++ b/@types/images.d.ts
@@ -6,7 +6,7 @@ interface StaticImageData {
     src: string;
     height: number;
     width: number;
-    // eslint-disable-next-line @typescript-eslint/naming-convention
+    // biome-ignore lint/style/useNamingConvention: <explanation>
     blurDataURL?: string;
 }
 
--- a/biome.json
+++ b/biome.json
@@ -19,8 +19,9 @@
       "complexity": {
         "all": false
       },
-      "performance": {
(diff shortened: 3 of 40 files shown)


MohammadxAli:
Great stuff 🙂✨ I first saw biome in an [open-source project (`viem`)](https://github.com/wevm/viem). It's new but I'm pretty sure it is well defined and tested.

> maybe we don't even need that much?

Agree!

There are 2 es-lint rules which I figured out we **ignore** a lot:

- `typescript-eslint/no-use-before-define`
- `import/export`

1. We usually define utils below the exported function from the file, which makes ESLint unhappy so we throw `typescript-eslint/no-use-before-define` at file-level.
2. We use TS namespaces a lot, so we ignore `import/export` rule for that case as well.

I believe for both of these cases we can disable these rules completely. What do you think?

oleg-semyonov:
Hey @MohammadxAli, thanks for your feedback!

I agree on the `no-use-before-define`, but we actually only ignore that for functions. This can be enabled as an option to the rule, since all function definitions are hoisted anyway. Not sure if Biome has that as an option, but we should likely update our ESLint config too.

As for `import/export`, I don't think that's the rule to disable, we only ignore the `no-namespace` rule. This rule is not enabled in the current Biome config, so we're covered.

e1himself (commented):
IDK, I get the joy of adopting new tools, but at this stage, it seems it will cause more pain when using it, compared to ESLint, even though it's faster to run.

So I'm not supporting this change (at the current stage of Biome evolution).

PR #1100, opened 2024-02-08
[CARE-4335] Feature - Only show categories that have public stories in the current locale

kudlajz:


e1himself (approved):

modules/Search/components/Hit.tsx
```diff
@@ -36,6 +36,7 @@ export function Hit({ hit }: Props) {
                     name,
                     slug,
                     description: null,
+                    public_stories_number: 0, // We do not have access to number of stories from Algolia
```
kudlajz, line 39: We don't really need this property in this code, so just adding it here to satisfy the type.

PR #1109, opened 2024-03-18
[DEV-12642] Feature - Display featured categories on homepage and allow filtering by them

kudlajz:
Screenshot 2024-03-18 at 18 14 16


modules/InfiniteStories/ui/CategoriesFilters/CategoriesFilters.tsx
```diff
@@ -0,0 +1,51 @@
...
+    activeCategory: Pick<Category, 'id'> | undefined;
+    categories: Category[];
+    className?: string;
+    locale: Locale.Code;
+}
+
+export function CategoriesFilters({ activeCategory, categories, className, locale }: Props) {
+    return (
+        <div className={className}>
+            <PageTitle className={styles.title} title="Latest stories" />
```
kudlajz, line 20: I'll have to request translations for these two strings, but for now we can deploy this as-is since users do not have access to featured categories in the app yet.

modules/Stories/Stories.tsx
```diff
@@ -1,27 +1,46 @@
...
     return (
         <InfiniteStories
+            key={categoryId}
+            category={categoryId ? { id: categoryId } : undefined}
+            categories={featuredCategories}
             newsroomName={languageSettings.company_information.name || newsroom.name}
             pageSize={pageSize}
-            initialStories={stories}
+            initialStories={
+                hasFeaturedCategories ? [...pinnedOrMostRecentStories, ...stories] : stories
```
kudlajz, line 42: This change is required to keep the pinned/most recent story static when you start filtering, so only the 2nd story and onwards changes.

The downside is that the pinned/most recent story can be potentially duplicated in the filtered stories below, but we couldn't think of a simple way to solve this that would not incorporate fiddling with the pagination.

modules/Stories/Stories.tsx
```diff
@@ -1,27 +1,46 @@
...
+
     const { stories, pagination } = await app().stories({
+        category: categoryId ? { id: categoryId } : undefined,
         limit: pageSize,
         locale: { code: localeCode },
     });
 
     return (
         <InfiniteStories
+            key={categoryId}
```
kudlajz, line 36: I've added the key prop here so the component is re-mounted and the `useInfiniteLoading` hook resets its state since we're providing different `initialStories`.

pestaa (approved):
LGTM

PR #1111, opened 2024-03-19
[DEV-12648] Implement featured categories on mobile menu

pestaa:
Screenshot_20240320_125128
Screenshot_20240320_124948


styles/variables/_typography.scss
```diff
@@ -1,7 +1,8 @@
 $font-weight-regular: 400;
 $font-weight-medium: 500;
 $font-weight-bold: 600;
-$font-weight-bolder: 900;
+$font-weight-bolder: 700;
+$font-weight-boldest: 900;
```
kudlajz, line 5: I think the 600 variant should be called `semi-bold` (same as in the app), 700 should be `bold` and 900 `bolder`, so it's consistent with the [CSS naming](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight).

components/CategoryImage/CategoryImage.tsx
```diff
@@ -0,0 +1,37 @@
...
+                objectFit="cover"
+                containerClassName={classNames(styles.imageContainer, className)}
+                className={styles.image}
+                sizes={getCardImageSizes(size)}
+            />
+        );
+    }
+
+    // TODO add fallback image
+    return null;
```
kudlajz, line 36: Are you planning on addressing this TODO? Right now if there's no image, the consumer code renders an empty link which creates some empty space.

components/CategoryImage/CategoryImage.tsx
```diff
@@ -0,0 +1,37 @@
+'use client';
```
kudlajz, line 1: This component doesn't seem to need `use client`? I've removed it and it still works.

components/CategoryImage/CategoryImage.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Image from '@prezly/uploadcare-image';
+import classNames from 'classnames';
+
+import { type CardSize, getCardImageSizes } from './lib';
+
+import styles from './CategoryImage.module.scss';
+
+type Props = {
+    category: Category;
+    translatedCategory: TranslatedCategory;
```
kudlajz, line 13: I think these props could be simpler since you only need the `image` and `alt` props, so maybe just add those instead of passing the whole `category` and `translatedCategory`?

modules/Header/ui/CategoriesNav/CategoriesNavMobile.tsx
```diff
@@ -0,0 +1,114 @@
...
+    return (
+        <>
+            <li className={navigationItemClassName}>
+                <Dropdown
+                    label={
+                        <FormattedMessage locale={localeCode} for={translations.categories.title} />
+                    }
+                    buttonClassName={navigationItemButtonClassName}
+                    withMobileDisplay
+                >
```
kudlajz, line 77: Should there be a bit more top padding on the dropdown?

Right now, the dropdown button is "touching" the first row of featured categories, so might be a bit difficult to hit?

Currently in production there's more space between the button and the categories themselves.

Here:
Screenshot 2024-03-20 at 13 00 46

Prod:
Screenshot 2024-03-20 at 13 25 04

kudlajz: I guess maybe it might not be such a big problem if we add the placeholder category image.

modules/Header/ui/CategoriesNav/CategoriesNavMobile.module.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.divider {
+    border: none;
+    border-top: 1px solid $color-borders;
+    margin: $spacing-2 0;
+}
+
+.featuredContainer {
+    display: flex !important;
```
kudlajz, line 21: Why using the important here? There's no other rule.

modules/Header/ui/CategoriesNav/CategoriesNavMobile.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    function isCategoryFeatured(translatedCategory: TranslatedCategory) {
+        return getCategory(translatedCategory).is_featured;
+    }
+
+    const featuredCategories = categories.filter(isCategoryFeatured);
+    const regularCategories = categories.filter((i) => !isCategoryFeatured(i));
+
+    if (showAllCategories) {
```
kudlajz, line 36: I'm not sure how to feel about this 🤔 Here it looks like the "Media galleries" is a title and the items below are media galleries, but those are actually categories.

localhost_3000_(iPhone 12 Pro)

localhost_3000_(iPhone 12 Pro) (1)

If I don't have any featured categories, the categories themselves look completely different from the other elements? 🤔 

localhost_3000_(iPhone 12 Pro) (2)

styles/variables/_typography.scss
pestaa: 2dd84afe

components/CategoryImage/CategoryImage.tsx
pestaa: b7322b5f

modules/Header/ui/CategoriesNav/CategoriesNavMobile.module.scss
pestaa: 70801e12

components/CategoryImage/CategoryImage.tsx
pestaa: What should we do? We haven't settled on a fallback image. Maybe we can reuse the one for the stories?

components/CategoryImage/CategoryImage.tsx
pestaa: Simpler in what sense? There's plenty of naming and chasing scalar props. I prefer to keep objects intact to match a higher level mental model.

kudlajz: I guess in this context it doesn't matter too much since it's a server component but in client components using scalar props makes it easier for React to check for value equality and therefore less chance of a re-render.

It's also easier to test components like this if you can just mock a single value instead of mocking the entire object.
Also also easier to extract the component into a generic component later, if it's not coupled to a specific object.

So I guess the tl;dr is that I like to keep the interface as simple as possible, if possible. This component could have quite easily be just called `Image` without the need for `Category` and it would still work, if the props were changed. Just saying.

So agree, it's a matter of preference 😉

components/CategoryImage/CategoryImage.tsx
pestaa: c1974274

kudlajz: We also have a fallback for when there's no newsroom logo 🙂

modules/Header/ui/CategoriesNav/CategoriesNavMobile.tsx
pestaa: a14afe38

modules/Header/ui/CategoriesNav/FeaturedCategory.tsx
```diff
@@ -0,0 +1,59 @@
...
+                className={styles.image}
+                href={{
+                    routeName: 'category',
+                    params: {
+                        slug: translatedCategory.slug,
+                        localeCode: translatedCategory.locale,
+                    },
+                }}
+            >
+                <CategoryImage
```
kudlajz, line 27: I forgot to report this but the images can become uneven:

Screenshot 2024-03-20 at 13 14 43

components/CategoryImage/CategoryImage.tsx
pestaa: Thanks for those points. Indeed my backend bias shows. Will keep them in mind for client components. :)

modules/Header/ui/CategoriesNav/FeaturedCategory.tsx
pestaa: 570bd301

kudlajz (approved):

PR #1116, opened 2024-03-21
[DEV-12659] Develop featured categories homepage block

pestaa:
This setup doesn't sell the concept too well, but it is what I cobbled together locally.

I'm not happy with the 5 items layout on desktop, curious to hear your thoughts.

Screenshot_20240322_021355
Screenshot_20240322_021334
Screenshot_20240322_021258


modules/FeaturedCategories/FeaturedCategories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface Props {
+    localeCode: Locale.Code;
+}
+
+export async function FeaturedCategories({ localeCode }: Props) {
+    const categories = await app().categories();
+    const translatedCategories = await app().translatedCategories(
+        localeCode,
+        categories.filter((i) => i.is_featured),
```
kudlajz, line 15: Probably should also filter out categories that have no stories like we do in [other places](https://github.com/prezly/theme-nextjs-bea/blob/main/modules/Header/Header.tsx#L22).

pestaa: 2d932f46

alpower:
1 cat will have to be full width I think. 

yeah I'm not sure about that desktop one either. I'd be tempted to go either way - stick to the 33%/max 50 of the size before, or make it 20% and have 5.

You're always going to have to suffer with an odd one out on it's own unless you either:
 - make it grow to be full width
 - count the number of featured categories, and conditionally change the style based on odds and evens (odds having a '33%/max 50' and evens having a % that gives an equal row number.

pestaa:
localhost_3000_

alpower (approved):

components/CategoryImage/CategoryImage.tsx
```diff
@@ -10,21 +10,20 @@ import { type CardSize, getCardImageSizes } from './lib';
 import styles from './CategoryImage.module.scss';
 
 type Props = {
-    category: Category;
-    translatedCategory: TranslatedCategory;
+    image: Category['image'];
+    name: TranslatedCategory['name'];
```
kudlajz, line 14: Technically this could have been just `ImageDetails` or whatever the type is and `string`, right? 😄

modules/FeaturedCategories/ui/CategoryImage.module.scss
```diff
@@ -0,0 +1,32 @@
...
+}
+
+.placeholder {
+    background-color: var(--prezly-accent-color);
+}
+
+.image, .placeholder {
+    @include border-radius-s;
+
+    height: 151px;
```
kudlajz, line 13: What an interesting number 🤔

alpower: He took it from my Figma (which I've since rounded down) but he seems to be particularly attached to it 😅 - I say we change it to 150 as is making me twitch @kudlajz

PR #1126, opened 2024-04-10
[DEV-12776] Fix - Story UUID is not passed into analytics context provider

kudlajz:
This is a bit of a hacky solution but I've spent quite some time on this already and couldn't find a better solution without majorly rethinking the whole analytics package.

By using the debounced function (which uses timeout), the `page` call is only triggered after the contexts are updated and components are re-rendered, making it so that all the data is up to date.

In order for this solution to work, it requires disabling automatic page tracking that I've prepared in https://github.com/prezly/analytics/pull/299.

### DIFF
--- a/modules/Broadcast/BroadcastStory.tsx
+++ b/modules/Broadcast/BroadcastStory.tsx
@@ -1,0 +1,50 @@
+'use client';
+
+import type { Story } from '@prezly/sdk';
+import type { ReactNode } from 'react';
+import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react';
+
+type Revoke = () => void;
+
+interface Context {
+    story: Story | null;
+    broadcast: (story: Story) => Revoke;
+}
+
+const context = createContext<Context>({
+    story: null,
+    broadcast() {
+        throw new Error(
+            'This functionality requires `BroadcastStoryProvider` mounted up the components tree.',
+        );
+    },
+});
+
+export function BroadcastStoryProvider(props: { children: ReactNode }) {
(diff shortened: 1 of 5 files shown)


modules/Broadcast/BroadcastStory.tsx
```diff
@@ -0,0 +1,50 @@
...
+const context = createContext<Context>({
+    story: null,
+    broadcast() {
+        throw new Error(
+            'This functionality requires `BroadcastStoryProvider` mounted up the components tree.',
+        );
+    },
+});
+
+export function BroadcastStoryProvider(props: { children: ReactNode }) {
```
kudlajz, line 23: Taking into consideration we are only interested in the story's UUID, do you think it would be better to only broadcast the UUID instead of the whole object? 🤔

MohammadxAli: I wouldn't, as long as it doesn't hurt which in this case it seems it doesn't.

Kinda reminds me of how we pass props to components. I realized the pattern of passing the whole entity is just cleaner, less verbose and more maintainable in the long run.

kudlajz: Yeah, I think it's okay too. The object should have a stable reference as well thanks to `useMemo` in case we need to use it as a dependency somewhere.

MohammadxAli (approved):

kudlajz:
OMG, I forgot to update the analytics package first. Will have to revert 🙈

kudlajz:
Okay, not reverting but will fix it in https://github.com/prezly/theme-nextjs-bea/pull/1127 before merging.

PR #1127, opened 2024-04-10
[DEV-12777] Feature - Pass media gallery UUID into analytics provider to allow better tracking

kudlajz:
Uses the same approach as in https://github.com/prezly/theme-nextjs-bea/pull/1126. Requires changes from https://github.com/prezly/analytics/pull/300.


MohammadxAli (approved):
👍

PR #1135, opened 2024-05-06
[CARE-3415] Feature - Use Uploadcare's image loader to optimize images

kudlajz:
I've replaced usages of `UploadcareImage` from our custom library with Uploadcare's own Nextjs image loader which generates all the required `srcset` URLs for us automatically.

There has been a significant image size reductions in most areas, mainly the logo and homepage (story cards) but of course depends on how big the original images were. Places like contact cards should see a big benefit as those were previously rendered in an unnecessary large dimensions.

Since Uploadcare's image loader is using `next/image` under the hood, it also comes with automatic lazy loading so places like homepage load even faster on smaller screens since the browser doesn't have to fetch all of them at once.


.env.example
```diff
@@ -18,3 +18,7 @@ ALGOLIA_API_KEY=
 # ALGOLIA_INDEX=
 
 # API_BASE_URL_OVERRIDE=
+
+# Uploadcare variables
+NEXT_PUBLIC_UPLOADCARE_PUBLIC_KEY=97775dfb0ac5a6446bce
+NEXT_PUBLIC_UPLOADCARE_CUSTOM_CDN_DOMAIN=cdn.uc.assets.prezly.com
```
kudlajz, line 24: Hey @camilb, I've added these two env variables but I believe we also need to add them to the `build-and-push.yml` file, right? So I assume they have to be defined as secrets first?

modules/Header/ui/Header.tsx
```diff
@@ -123,13 +127,14 @@ export function Header({
...
-                                objectFit="contain"
-                                imageDetails={newsroom.newsroom_logo}
+                        {newsroomLogo && (
+                            <UploadcareImage
+                                style={{ objectFit: 'contain' }}
+                                src={newsroomLogo.cdnUrl}
                                 alt="" // This is a presentation image, the link has text inside <h1>, no need to have it twice. See [DEV-12311].
                                 className={styles.logo}
+                                width={320}
+                                height={56}
```
alpower, line 134: @kudlajz I'm just starting a spike on looking at how we can better accommodate peoples different sized logos.

Looking at the current Bea, we set an equivalent `max-width` and `max-height` of these values. Does this component result in the same outcome? i.e. can I assume that this will behave similarly as current, for the purposes of thinking what I do next?

CleanShot 2024-05-07 at 10 56 16@2x

kudlajz: Yes, the dimensions are still the same, I've just set the width and height so the component fetches the image in 320px width (twice that for Retina) and not more.

alpower: Cool -  thanks for confirming! Just checking that that width and height weren't now hardcoded as the width and height of the output or anything like that - I'll assume the CSS is as is current for this PR in terms of how we treat logos, so this is v much just changing how we fetch the logo more optimally 👍

kudlajz: > so this is v much just changing how we fetch the logo more optimally

Yes 🙂

alpower (approved):
Will let a dev properly check the code, but at a glance this looks ace! Thanks.

MohammadxAli (approved):
👍

PR #1144, opened 2024-05-14
[CARE-4999] Feature - Update CSP header to allow iframing

kudlajz:
Updated the theme kit to include changes from https://github.com/prezly/theme-kit-js/pull/682.


camilb (approved):
👍

PR #1145, opened 2024-05-15
[CARE-5001] Fix - Preview parameters are not applied correctly

kudlajz:
- search params parsing was fixed
- implemented new hook called `useThemeSettingsWithPreview` which takes the actual theme settings and merges the preview settings into it
- replaced usages of `useThemeSettings` with `useThemeSettingsWithPreview` so preview reflects the user's choices in the app


modules/Head/components/DynamicPreviewBranding.tsx
```diff
@@ -16,15 +15,12 @@ interface Props {
 }
 
 export function DynamicPreviewBranding({ settings }: Props) {
-    const searchParams = JSON.stringify(useSearchParams());
```
kudlajz, line 19: So, the main problem was that calling `JSON.stringify` was stringifying the result of `useSearchParams` to an empty object even if there were parameters present.

MohammadxAli (approved):
👍

PR #1147, opened 2024-05-16
[CARE-5022] Feature - Create a mask when preview=true query parameter is present

kudlajz:
Implemented a component that creates a fixed overlay covering the entire page, preventing the user from clicking anything, but still allow scrolling.

This functionality only works when `preview=true` query parameter is present and is intended to work with our live theme preview functionality.


alpower (approved):
Love seeing my ideas come to life! 🤓 😍

PR #1154, opened 2024-06-03
[CARE-5202] Feature - Hide cookie bar if mask parameter is present

kudlajz:
When live previewing the theme, we use the `mask` parameter to prevent clicking inside the page, but cookie consent bar could still appear.

With this change, the cookie bar will not be rendered.

### DIFF
--- a/hooks/useMaskParam.ts
+++ b/hooks/useMaskParam.ts
@@ -1,0 +1,21 @@
+'use client';
+
+import { useSearchParams } from 'next/navigation';
+import { useMemo } from 'react';
+
+export function useMaskParam(): boolean {
+    const searchParams = useSearchParams();
+    const mask = searchParams.get('mask');
+
+    return useMemo(() => {
+        if (!mask) {
+            return false;
+        }
+
+        try {
+            return Boolean(JSON.parse(mask));
+        } catch (_) {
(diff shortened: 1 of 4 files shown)


hooks/useMaskParam.ts
```diff
@@ -0,0 +1,21 @@
+'use client';
+
+import { useSearchParams } from 'next/navigation';
+import { useMemo } from 'react';
+
+export function useMaskParam(): boolean {
```
MohammadxAli, line 6: I guess if we call this `useIsMaskParamAvailable` then it makes more sense based on the hook's output.

MohammadxAli (approved):
👍

hooks/useMaskParam.ts
kudlajz: I'm not sure the suggested name is better 🤔 Since the hook returns the parsed `mask` param, I think the name is fine? 🤔

fgyimah (approved):
👍🏾

hooks/useMaskParam.ts
MohammadxAli: Ah. I see! My bad :) so the value is either `true` or `false`. Got it :)

PR #1155, opened 2024-06-04
[CARE-5201] Feature - Allow communication between the theme and the parent

kudlajz:
We need to be able to preserve the scroll position when live previewing the theme, but since it's not possible to access the scroll position of the iframe, we need this extra code to provide the scroll position when scrolling happens.

Similarly, we also need to listen to the messages from the parent so we can restore the scroll position after the iframe is reloaded.


fgyimah (approved):
👍🏾

PR #1163, opened 2024-06-27
[CARE-5635] Feature - Support new preview options

kudlajz:
Added support for new settings and also previewing them
- logo
- main site link
- toggling share icons on story page
- story header image placement (above/below)

I've also optimized some of the preview parameters to be parsed and applied in the server components instead of on the client, so there's less flash of content/content changing on render.


app/[localeCode]/(story)/secret/[uuid]/page.tsx
```diff
@@ -12,6 +13,7 @@ interface Props {
         localeCode: Locale.Code;
         uuid: StoryRef['uuid']; // story secret_uuid
     };
+    searchParams: Record<string, string>;
```
Comment (e1himself) [inline app/[localeCode]/(story)/secret/[uuid]/page.tsx:16]: I wonder why `searchParams` is generic here, and is more specific in the definition above.

modules/Footer/ui/Footer.tsx
```diff
@@ -1,21 +1,36 @@
...
+    const searchParams = useSearchParams();
+
+    const isWhiteLabeled = useMemo(() => {
+        const isWhiteLabeledPreview = searchParams.get('is_white_labeled');
+        if (isWhiteLabeledPreview) {
+            return parseBoolean(isWhiteLabeledPreview);
+        }
+
+        return props.isWhiteLabeled;
+    }, [props.isWhiteLabeled, searchParams]);
```
e1himself, line 26: Why `useMemo()`? Is this an expensive operation? :)

modules/Head/components/DynamicPreviewBranding.tsx
```diff
@@ -17,7 +17,18 @@ interface Props {
...
+        () =>
+            Array.from(searchParams.entries()).reduce(
+                (result, [key, value]) => ({
+                    ...result,
+                    [key]: value,
+                }),
+                {},
+            ),
+        [searchParams],
+    );
```
e1himself, line 30: Can this be moved into `parsePreviewSearchParams()`?

modules/Header/ui/Header.tsx
```diff
@@ -198,6 +231,19 @@ export function Header({
                                     marginTop={measurement?.height}
                                     translatedCategories={translatedCategories}
                                 />
+                                {mainSiteUrl && (
+                                    <li className={styles.navigationItem}>
+                                        <ButtonLink
+                                            href="https://www.prezly.com"
```
e1himself, line 237: > href="https://www.prezly.com"

Is this not a mistake?

modules/Header/ui/Header.tsx
```diff
@@ -218,3 +264,8 @@ export function Header({
         </header>
     );
 }
+
+function humanizeUrl(url: URL) {
+    const string = url.hostname.replace('www.', '');
```
e1himself, line 269: What if the domain name has `www` in it? Like `holycowww.com`? :) 

```suggestion
    const string = url.hostname.replace(/^www\./, '');
```

modules/Story/Story.tsx
```diff
@@ -32,10 +38,13 @@ export async function Story({ story }: Props) {
         <article className={styles.story}>
             <div className={styles.container}>
                 <Embargo story={story} />
+                {withHeaderImage === 'above' && headerImageDocument && (
+                    <HeaderImageRenderer nodes={headerImageDocument} />
+                )}
```
e1himself, line 43: I wonder if this cannot become a part of the `<HeaderRenderer>`. Sounds about right. Not insisting though :)

e1himself (approved):

app/[localeCode]/(story)/secret/[uuid]/page.tsx
Comment (kudlajz) [reply app/[localeCode]/(story)/secret/[uuid]/page.tsx:16]: I think I've made it more specific when I was first implementing it and then I realised that it doesn't really matter to the "route" to be too specific here, because we pass the `searchParams` to the parsing function anyway, which checks and picks the keys it expects.

So I think I can use the generic type everywhere.

modules/Head/components/DynamicPreviewBranding.tsx
kudlajz: Maybe.. I wasn't sure if it's the best idea to make the `parsePreviewSearchParams` work with both plain objects as well as `URLSearchParams` interface since most of the time the input argument is a plain object.

We're only passing the `URLSearchParams` once in this place, so I thought it's okay to have the code here.

modules/Header/ui/Header.tsx
kudlajz: It is 🙈

modules/Header/ui/Header.tsx
kudlajz: You're right, I apologize for all the cowwws out there, thanks! 😆

modules/Footer/ui/Footer.tsx
kudlajz: No, it's not 🙂 Removed.

PR #1176, opened 2024-08-13
[DEV-13294] Support new color settings

kudlajz:


components/RichText/Quote.module.scss
```diff
@@ -18,16 +18,15 @@ $content-padding: $spacing-3 + $spacing-4;
...
     position: relative;
 
     &::before {
         content: "";
         position: absolute;
         top: 50%;
-        background-color: $color-base-400;
-        width: 2px;
+        background-color: var(--prezly-border-color-secondary);
+        width: 4px;
```
alpower, line 29: Checking we meant to increase the quote border width.

components/StoryCards/StoryCard.module.scss
```diff
@@ -27,7 +22,7 @@ $spacing-card: $spacing-4 - $spacing-half; // 1.25rem = 20px That's the only hac
     display: block;
     overflow: hidden;
     text-decoration: none;
-    border: 1px solid $color-borders;
+    border: 1px solid var(--prezly-border-color);
```
alpower, line 25: a touch too strong? should we drop the opacity of this border a little?
CleanShot 2024-08-19 at 12 42 39@2x

alpower (approved):
Looks good - two minor comments.

components/StoryCards/StoryCard.module.scss
kudlajz: I can barely notice the border to be honest.

components/RichText/Quote.module.scss
kudlajz: But I did notice that I forgot to add the background color to the Quote.

fgyimah (approved):
👍🏾

components/RichText/Quote.module.scss
kudlajz: Wait, I didn't actually, because in Figma, quote is transparent whereas in production it's using something that would be a "bg secondary" in the new variables.. so not sure which one to use.

kudlajz: So either we go with the new transparent + 4px width or keep 2px width but use the secondary bg color.

alpower: hmm - I wonder if quotes can be included in anything? I think it's OK to use transparent TBH as they should be always on something with a background color.

alpower: Quote width must be from Hugo's new designs - I'm happy w that then. No further comments.

PR #1177, opened 2024-08-19
[DEV-13354] Add support for more fonts

kudlajz:
- added support for more fonts
- some fonts now have a secondary font
  - in this case, primary font is used only for headings


fgyimah (commented):
We are missing the support for the secondary font `Alegreya sans`

fgyimah (approved):
👍🏾

alpower (approved):
Nice job!

PR #1206, opened 2025-01-09
[DEV-19082] Feature - Implement new error page for Bea theme

fgyimah:
image


app/[localeCode]/error.tsx
```diff
@@ -0,0 +1,46 @@
...
+'use client';
+
+import Illustration from '@/public/images/error.svg';
+
+import 'modern-normalize/modern-normalize.css';
+import '@/styles/styles.globals.scss';
+
+import styles from './error.module.scss';
+
+export default function Error() {
```
Comment (kudlajz) [inline app/[localeCode]/error.tsx:12]: Hmm.. I wonder if it's possible to only display the error page and not the rest like the header etc.? 🤔

app/[localeCode]/error.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return (
+        <div className={styles.container}>
+            <div className={styles.wrapper}>
+                <div className={styles.illustration}>
+                    <Illustration className={styles.svg} />
+                </div>
+                <div className={styles.content}>
+                    <h1 className={styles.title}>We’re sorry, this newsroom couldn’t load</h1>
```
Comment (kudlajz) [inline app/[localeCode]/error.tsx:24]: We do not use the word `newsroom` anymore.

app/[localeCode]/error.tsx
Comment (fgyimah) [reply app/[localeCode]/error.tsx:12]: It uses the folder layout component, not sure of a way to bypass that yet

Comment (kudlajz) [reply app/[localeCode]/error.tsx:12]: What about this? https://nextjs.org/docs/pages/building-your-application/routing/custom-error#customizing-the-500-page

Comment (kudlajz) [reply app/[localeCode]/error.tsx:12]: What if you move the `error.tsx` file directly into `app` folder? Does that work or not?

Comment (fgyimah) [reply app/[localeCode]/error.tsx:12]: Tried but that also does not work

Comment (kudlajz) [reply app/[localeCode]/error.tsx:12]: Hmm, okay. How did you simulate the error for the screenshot? I'm wondering how it's going to look if all of the API request return a 500 for example.

Comment (fgyimah) [reply app/[localeCode]/error.tsx:12]: I did this:

image

Comment (kudlajz) [reply app/[localeCode]/error.tsx:12]: Right, but that is an error on a specific page. Can you throw the error somewhere else, maybe in the main layout component?

I just want to see how it works if the error happens outside of the actual page (before the app can be even rendered).

Comment (fgyimah) [reply app/[localeCode]/error.tsx:12]: It doesn't work, fixing with this: https://nextjs.org/docs/app/api-reference/file-conventions/error#global-error

modules/Error/Error.tsx
```diff
@@ -0,0 +1,56 @@
+'use client';
```
kudlajz, line 1: Do all these files need to be client components?

The `window.location.reload` can technically be replaced just by using empty `href=""` on the refresh link, that should trigger a page reload.

Would be much better to have this server side rendered.

fgyimah: All the error pages should be client components according to the docs

kudlajz (approved):

PR #1212, opened 2025-01-22
[DEV-19242] Feature - Sharing & Downloading

yuriyyakym:


modules/Story/Share/utils/htmlToText.ts
```diff
@@ -0,0 +1,15 @@
+'use server';
```
yuriyyakym, line 1: In dev environment (unminified scripts), having it as a server action saves ~260kb. Not sure how much would it be in prod though.

modules/Story/Share/Share.tsx
```diff
@@ -0,0 +1,145 @@
...
+        link.click();
+        document.body.removeChild(link);
+    }
+
+    return (
+        <>
+            <Divider />
+            <div>
+                <h2>Share</h2>
+
```
pestaa, line 79: This'll be out of place on non-English sites. We'll probably have to add translations.

modules/Story/Share/utils/getStoryPdfUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+    prezlyClient.api
+        .get(`${routing.storiesUrl}/${uuid}`, {
+            headers: {
+                Accept: 'application/pdf',
+            },
+        })
+        .then(console.log)
+        .catch(console.log);
+
+    // TODO: return a proper PDF url
```
pestaa, line 20: Can I help here? It's supposed to be the Location header of the response.

yuriyyakym: It's done, excuse me, pushed to a wrong branch

modules/Story/Share/Share.tsx
yuriyyakym: I would need some assistance with that

pestaa (approved):

PR #1217, opened 2025-02-05
[DEV-19367][PRD-1243] Feature - Story - Latest stories section

yuriyyakym:
image

### DIFF
--- a/app/[localeCode]/(story)/[slug]/page.tsx
+++ b/app/[localeCode]/(story)/[slug]/page.tsx
@@ -48,7 +48,7 @@ export default async function StoryPage({ params, searchParams }: Props) {
                 story={story}
                 showDate={themeSettings.show_date}
                 withHeaderImage={themeSettings.header_image_placement}
-                relatedStories={relatedStories}
+                relatedStories={themeSettings.show_read_more ? relatedStories : []}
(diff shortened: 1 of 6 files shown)


app/[localeCode]/(story)/[slug]/page.tsx
```diff
@@ -16,12 +16,18 @@ interface Props {
...
+    const { localeCode, slug } = params;
 
     const story = await app().story({ slug });
     if (!story) notFound();
 
-    return { story };
+    const { stories: relatedStories } = await app().stories({
+        limit: 3,
+        locale: localeCode,
+        query: JSON.stringify({ slug: { $ne: slug } }),
```
Comment (pestaa) [inline app/[localeCode]/(story)/[slug]/page.tsx:27]: Perhaps specify a sort explicitly? I imagine it's chronological by default, so not needed necessarily.

pestaa (approved):

app/[localeCode]/(story)/[slug]/page.tsx
Comment (yuriyyakym) [reply app/[localeCode]/(story)/[slug]/page.tsx:27]: The requirement was to do it as on homepage, and there we do not apply sorting:
https://github.com/prezly/theme-nextjs-bea/blob/main/modules/Stories/Stories.tsx

PR #1245, opened 2025-02-28
[DEV-18744] Feature - Update attachment presentation

e1himself:
See https://linear.app/prezly/issue/DEV-18744/attachment-embeds-look-bad-in-campaigns#comment-e25c5a7c

Screenshot From 2025-02-28 17-22-54

### DIFF
--- a/components/ContentRenderer/components/Attachment/utils.ts
+++ b/components/ContentRenderer/components/Attachment/utils.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
 // TODO: Should this be translated?
 const UNITS = ['bytes', 'Kb', 'Mb', 'Gb'];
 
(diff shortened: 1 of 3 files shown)


components/ContentRenderer/components/Attachment/utils.ts
```diff
@@ -13,3 +15,96 @@ export function formatBytes(bytes: number | string): string {
...
+ * Except unrecognized "octet-stream" and "x-download" fallbacks.
+ *
+ * ```sql
+ * SELECT node->'file'->>'mime_type', COUNT(*) FROM press_release, jsonb_array_elements(slate_content_json->'children') node
+ * WHERE node->>'type' = 'attachment'
+ * GROUP by 1
+ * ORDER by 2 DESC;
+ * ```
+ */
+const MAPPING: Record<string, string | undefined> = {
```
yuriyyakym, line 30: Maybe it's better to use dedicated [mime-type parsing library ](https://www.npmjs.com/package/mime-types). So that if new mime type appears in some story it would work without updating the theme.
This is used on back-end side anyways, so we should not be concerned about lib size (I haven't checked it)

yuriyyakym (approved):

components/ContentRenderer/components/Attachment/utils.ts
e1himself: >  This is used on back-end side anyways, so we should not be concerned about lib size (I haven't checked it)

Nope. 

image

I also think it's worth keeping it in-sync with the editor and email renderer, which do use this function.

PR #1252, opened 2025-03-03
[DEV-19722] Privacy Policy link below subscribe

pestaa:


modules/SubscribeForm/ui/SubscribeForm.tsx
```diff
@@ -136,10 +138,7 @@ export function SubscribeForm({ newsroom }: Props) {
                             ),
                             privacyPolicyLink: (
                                 <a
-                                    href={
-                                        newsroom.custom_privacy_policy_link ??
-                                        'https://www.prezly.com/privacy-policy'
-                                    }
+                                    href={PRIVACY_POLICY_PAGE}
```
kudlajz, line 141: I don't think this his how we do links. It should use the `Link` component from `@/components` and provide the `routeName`.

I mean, it works for now but it's gonna break if the URL ever changes.

kudlajz (approved):

modules/SubscribeForm/ui/SubscribeForm.tsx
pestaa: Followed https://github.com/prezly/theme-nextjs-bea/blob/5876756d630e4617c6d16437864c91243e4f284d/modules/CookieConsent/components/VanillaCookieConsent/VanillaCookieConsent.tsx#L24

But I guess it had other reasons to do it that way.

kudlajz: Probably because of passing the HTML as a string, but in your case should be fine to use a component.

pestaa: d6e76fe6 :+1:

PR #1254, opened 2025-03-03
[DEV-18744] Fix usage of a non-direct dependency

e1himself:
The type is re-exported by `@prezly/sdk`


pestaa (approved):
Thank you! I should've thought of this.

PR #1280, opened 2025-04-09
[DEV-20109] Fix - wide button looking off

fgyimah:
image

### DIFF
--- a/components/ContentRenderer/ContentRenderer.module.scss
+++ b/components/ContentRenderer/ContentRenderer.module.scss
@@ -144,6 +144,10 @@ $avatar-size: 60px;
                     border-color: var(--prezly-border-color-secondary);
                 }
             }
+
+            &--wide {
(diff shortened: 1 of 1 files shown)


components/ContentRenderer/ContentRenderer.module.scss
```diff
@@ -144,6 +144,10 @@ $avatar-size: 60px;
                     border-color: var(--prezly-border-color-secondary);
                 }
             }
+
+            &--wide {
+                width: 100% !important;
+            }
```
kudlajz, line 150: I think we should fix this in the content renderer itself.

kudlajz (approved):

components/ContentRenderer/ContentRenderer.module.scss
fgyimah: I'll prepare a separate PR in the renderer

PR #1295, opened 2025-04-30
[DEV-20302] Feature - Support for hub rooms

kudlajz:


src/modules/InfiniteHubStories/InfiniteHubStories.module.scss
```diff
@@ -0,0 +1,88 @@
...
+        &[data-count="8"] {
+            grid-template-columns: repeat(4, 1fr);
+        }
+    }
+
+    @include tablet-only {
+        grid-template-columns: repeat(3, 1fr);
+
+        &[data-count="1"] {
+            grid-template-columns: 373.33px;
```
yuriyyakym, line 40: 🤔  very precise. where does this number come from?

yuriyyakym (approved):

src/modules/InfiniteHubStories/InfiniteHubStories.module.scss
kudlajz: It's 1120px / 3. I'm not sure why @iamhugofelix used this container width because our containers are 1200px wide, so maybe we can consider using 1200/3 which is 400px. What do you think @iamhugofelix ?

app/api/hub-stories/route.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+
+    const { pagination, stories } = await app().client.stories.search({
+        offset,
+        limit,
+        include: ['thumbnail_image'],
+        sortOrder: SortOrder.desc('published_at'),
+        query: {
+            $and: [
+                { [`locale`]: { $in: [locale] } },
```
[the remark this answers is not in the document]
kudlajz: Yes, we always pass a locale when calling the API.

app/api/hub-stories/route.ts
```diff
@@ -0,0 +1,46 @@
...
+                { [`status`]: { $in: [Story.Status.PUBLISHED] } },
```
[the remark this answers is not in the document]
kudlajz: `$eq` doesn't seem to be supported, see https://developers.prezly.com/docs/api/47d07daed7b1d-search-stories.

I'm not sure why both `locale` and `status` are using `$in` operators though as I don't think they support multiple locales/statuses, but hey.

src/modules/InfiniteHubStories/InfiniteHubStories.module.scss
kudlajz: Now that I think about it, I guess your initial width makes sense as it's one column out of three, centered, right?

So it's 1200-2*40 (gap) / 3.

So let's keep it, I'll just add the calculation in the CSS instead of having the explicit number there.

PR #1298, opened 2025-05-16
Fix - Use newsroom UUID to check if story is external

kudlajz:


src/components/StoryCards/StoryCard.tsx
```diff
@@ -52,7 +52,8 @@ export function StoryCard({
     const hasCategories = translatedCategories.length > 0;
     const HeadingTag = size === 'small' ? 'h3' : 'h2';
 
-    const isExternalStory = window.location.hostname !== getHostname(url);
+    const isExternalStory =
```
e1himself, line 55: Can we check newsroom ID rather than URLs?

PR #1300, opened 2025-05-20
[DEV-20511][DEV-20510][DEV-20508] Fix - Various tweaks for hub rooms and search

kudlajz:
- improved search page
  - hide number of results if it's zero
  - hide filters if there's no facets to filter with
- added "Go to site X" title to site logos
- added scale effect to site logos on hover


fgyimah (approved):
👍🏾

PR #1304, opened 2025-05-22
[DEV-20547] Feature - Add support for search in hub sites

kudlajz:
This PR prepares hub sites for search functionality. We won't be enabling the search just yet because we need to update the Meilisearch index to provide story URLs as we can't construct story URLs in a reliable way (due to some sites having custom sub paths, routing etc.).

For this to work, we have to do some larger changes across the theme:
- correctly link to the external story in both search listings as well as regular listings
- correctly link to the external categories on external stories
- for external stories without thumbnail image, we have to use that site's logo, name and colors for placeholder


src/components/StoryCards/StoryCard.tsx
```diff
@@ -46,15 +49,14 @@ export function StoryCard({
...
-    url,
     variant = 'default',
     withStaticImage = false,
 }: Props) {
     const hasCategories = translatedCategories.length > 0;
     const HeadingTag = size === 'small' ? 'h3' : 'h2';
 
     const href = isExternal
-        ? url
+        ? isExternal.storyUrl
```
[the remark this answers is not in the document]
kudlajz: Maybe `withExternalUrl`? 🤔

src/components/StoryCards/StoryCard.tsx
```diff
@@ -13,9 +13,11 @@ import styles from './StoryCard.module.scss';
 
 type Props = {
     className?: string;
+    fallback: StoryImage.Props['fallback'];
     forceAspectRatio?: boolean;
-    isExternal: boolean;
+    isExternal: ExternalStoryUrl;
```
e1himself, line 18: It's a bit weird to see this boolean-like `is-` property used for data. 

As a suggestion, what do you think about naming it just `external` (without the `is-` prefix)?

e1himself (approved):
:+1:

src/components/StoryCards/StoryCard.tsx
kudlajz: Sure, sounds good to me, I'll change it.

PR #1328, opened 2025-06-20
[DEV-20811] Fix - Hide "Latest stories" H1 on homepage in certain conditions

kudlajz:
We need to render the "Latest stories" H1 to be compliant with the WCAG guidelines, however it's not needed visually for most users.

So this is how it works now:
- for non-hub sites
  - we only show the title if there's any featured categories
  - otherwise we only show it for screen readers
- for hub sites
  - we only show the title if there's any included stories
  - otherwise we only show it for screen readers

I've also fixed two more things:
1. removed extra bottom margin on the site logos in hub sites if there's no included stories
2. replaced the untranslated "Latest News" string with the translated "Latest stories"


fgyimah (approved):
👍

PR #1341, opened 2025-07-28
[DEV-21191] Fix - Do not render extra page title on category/tag pages

kudlajz:
Since category/tag pages already render their own title, we don't want to render the "Latest stories" in that case as that's reserved only for homepage.

### DIFF
--- a/src/modules/InfiniteStories/StoriesList.tsx
+++ b/src/modules/InfiniteStories/StoriesList.tsx
@@ -35,6 +35,7 @@ type Props = {
     stories: ListStory[];
     storyCardVariant: ThemeSettings['story_card_variant'];
     withEmptyState?: boolean;
+    withPageTitle?: boolean;
 };
 
 export function StoriesList({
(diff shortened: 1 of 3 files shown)


src/modules/InfiniteStories/StoriesList.tsx
```diff
@@ -113,14 +115,16 @@ export function StoriesList({
...
-            <PageTitle
-                className={classNames(styles.pageTitle, {
-                    // We want to hide the page title for regular users, but keep it
-                    // for the screen readers.
-                    [styles.aria]: isCategoryList ? !hasStories : !hasCategories,
-                })}
-                title={formatMessage(translations.homepage.latestStories)}
-            />
+            {withPageTitle && (
+                <PageTitle
```
yuriyyakym, line 119: If it's only used in Stories module, it can be rendered directly there🤔

yuriyyakym (approved):

PR #1346, opened 2025-09-05
[DEV-21506] Fix - Untranslated strings

e1himself:
Uses new strings from https://github.com/prezly/theme-kit-js/releases/tag/v10.4.0

### DIFF
--- a/src/modules/Story/Share/Share.tsx
+++ b/src/modules/Story/Share/Share.tsx
@@ -95,7 +95,7 @@ export function Share({
         <>
             <Divider />
             <div>
-                <h2>Share</h2>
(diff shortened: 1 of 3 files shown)


src/modules/Story/Share/Share.tsx
```diff
@@ -171,3 +171,7 @@ export function Share({
         </>
     );
 }
+
+function trim(text: string) {
+    return text.replace(/^[\s:]+|[\s:]+$/g, '');
+}
```
e1himself, line 177: There's already a string for `Share`, but it has colon in it: `Share:`

kudlajz (approved):

fgyimah (approved):

PR #1354, opened 2025-09-24
[DEV-21711] Feature - Use biome

yuriyyakym:


src/modules/Head/components/DynamicPreviewBranding.tsx
```diff
@@ -34,8 +34,7 @@ export function DynamicPreviewBranding({ settings }: Props) {
 
     useEffect(() => {
         setPreviewSettings(parsedPreviewSettings);
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [JSON.stringify(parsedPreviewSettings), setPreviewSettings]);
+    }, [setPreviewSettings, parsedPreviewSettings]);
```
kudlajz, line 37: `parsedPreviewSettings` is an object that is not memoized so this effect will keep triggering. It was correct before using the `JSON.stringify`.

kudlajz (approved):

PR #1355, opened 2025-09-30
[DEV-21762] Fix - Tweak hub site logo layout

kudlajz:
Switched around the conditions so now the layout is correct when there's 1, 2 or more than 9 logos. For everything in between it works the same.

I've also tweaked the layout for 1 and 2 columns so the logos are smaller and centered.


fgyimah (approved):
👍🏾

PR #1356, opened 2025-09-30
Fix - Provide pathname as dependency

kudlajz:
This effect got broken after migrating to Biome because without the `pathname` as a dependency, the effect did not trigger on navigation.

Unused `pathname` also made the Biome checks fail, preventing deployments.


fgyimah (approved):
👍🏾

PR #1363, opened 2025-10-03
Remove the underline also in the highlighted story card

JesseWynants:
Forgot to update the code also in the HighlightedStoryCard


fgyimah (approved):
👍🏾

PR #1436, opened 2026-04-24
[DEV-22111] Add reverse-proxy subpath support

JesseWynants:
Adds opt-in per-request base path handling for newsrooms served behind a reverse proxy at a subdirectory. Config comes via the existing X-Prezly-Env pipeline as two optional variables:

- THEME_BASE_PATH: flat prefix at root (e.g. /presscentre)
- THEME_LOCALE_SUBDIR: subdir after the locale slug, per-locale or '*' wildcard (e.g. *:news for /<locale>/news/*)

Server routing() and client useRouting() wrap generateUrl so every emitted URL (pages, metadata, sitemap, Link, search, broadcast) picks up the prefix automatically. Middleware parses X-Prezly-Env inline (Edge-safe, no next/headers) and strips the prefix before IntlMiddleware runs, then re-applies it to any redirect Location header so the browser stays on the user-visible URL. BasePathConfig is serialized through a BasePathProvider so the client wrap sees the same value.

Unset vars are a no-op; standard Bea is unchanged.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,3 +22,16 @@ MEILISEARCH_API_KEY=
 # Uploadcare variables
 NEXT_PUBLIC_UPLOADCARE_PUBLIC_KEY=97775dfb0ac5a6446bce
 NEXT_PUBLIC_UPLOADCARE_CUSTOM_CDN_DOMAIN=cdn.uc.assets.prezly.com
+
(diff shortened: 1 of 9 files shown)


camilb:
CleanShot 2026-04-24 at 14 48 08@2x
@JesseWynants AI review found this

JesseWynants:
ah yeah, good catch

camilb (approved):

PR #1438, opened 2026-04-28
[DEV-22375] Implement anchor links in themes

fgyimah:
- Adds id attributes to rendered h3/h4 heading elements in story content, using slugified versions of the heading text. This enables anchor linking directly to specific sections within a story.

### DIFF
--- a/src/components/RichText/Heading.tsx
+++ b/src/components/RichText/Heading.tsx
@@ -27,7 +27,7 @@ export function Heading({ node, children }: Props) {
         [styles.alignJustify]: node.align === TextAlignment.JUSTIFY,
     });
 
-    const id = slugifyHeading(extractText(node.children));
+    const id = `header-${slugifyHeading(extractText(node.children))}`;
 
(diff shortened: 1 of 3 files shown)


src/components/RichText/Heading.tsx
```diff
@@ -19,9 +27,19 @@ export function Heading({ node, children }: Props) {
         [styles.alignJustify]: node.align === TextAlignment.JUSTIFY,
     });
 
+    const id = slugifyHeading(extractText(node.children));
```
e1himself: What will happen if this generates and `id` that is already present on the page? 
Do you think it makes sense to prefix it with `header-...` or something?

e1himself (approved):

src/components/RichText/Heading.tsx
fgyimah: Agree, will update

fgyimah: https://github.com/prezly/theme-nextjs-bea/pull/1438/commits/b22b1504abf8d769a310176a2bb38dae6108bc24

PR #1439, opened 2026-04-30
[DEV-22377] Feature - Support new Meilisearch index

fgyimah:
Based on prezly/theme-kit-js#1025

- Upgraded `@prezly/theme-kit-nextjs` to `10.6.3-beta.0`, which adds the `IndexedStorySection` type with `section_title`, `section_subtitle`, and `url` fields
- Added `hash` support to the `Link` component's route object href, enabling anchor navigation via internal routes
- Added an `anchor` prop to `StoryCard` that appends to the story URL for both internal and external links
- Updated the search results page and search widget to use `IndexedStorySection`, compute an anchor from `section_title` (falling back to `section_subtitle`), and link directly to the matched section using `#header-{slug}`
- Fixed external cross-newsroom story URLs in both search results and search widget to use the `url` field from the index instead of an empty string
- Changed the default MeiliSearch index to `public_stories_v2`

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
         "@prezly/content-renderer-react-js": "0.44.0",
         "@prezly/sdk": "23.17.0",
         "@prezly/story-content-format": "0.70.0",
-        "@prezly/theme-kit-nextjs": "10.6.2",
+        "@prezly/theme-kit-nextjs": "10.6.3-beta.0",
         "@prezly/uploadcare": "2.6.0",
         "@prezly/uploadcare-image": "0.3.2",
         "@react-hookz/web": "14.7.1",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
         "@prezly/content-renderer-react-js": "0.44.0",
         "@prezly/sdk": "23.17.0",
         "@prezly/story-content-format": "0.70.0",
-        "@prezly/theme-kit-nextjs": "10.6.2",
+        "@prezly/theme-kit-nextjs": "10.6.3-beta.0",
```
fgyimah: TODO: Upgrade to a more stable version after prezly/theme-kit-js#1025 is merged

fgyimah: https://github.com/prezly/theme-nextjs-bea/pull/1439/commits/02b3b8c2b991bd81385dfba7d9913b745944955b

digitalbase (approved):
Looks good. Make sure to test this after deployment

PR #1443, opened 2026-05-18
DEV-23074: Add hidecookie query param for preview logic

boki-prezly:
Add `?hidecookie=true` query param support that hides the cookie consent banner (both Vanilla and OneTrust) without showing the preview bar. This allows in-app iframe previews to use `?mask=true&hidecookie=true` to block clicks and hide the cookie banner, while keeping the preview bar hidden.


fgyimah (approved):
Good work!

PR #1447, opened 2026-05-25
chore: add font antialiasing to body

iamhugofelix:
Applies -webkit-font-smoothing: antialiased and -moz-osx-font-smoothing: grayscale to the body element, enabling grayscale anti-aliasing for text rendering on macOS across all browsers. 

**This produces crisper, lighter-weight text compared to the default subpixel rendering.**

### DIFF
--- a/src/styles/_typography.scss
+++ b/src/styles/_typography.scss
@@ -6,6 +6,8 @@ body {
     color: var(--prezly-text-color);
     word-break: break-word;
     background: var(--prezly-background-color);
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
 }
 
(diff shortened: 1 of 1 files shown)


fgyimah (approved):
Image

harmasz (approved):
One quick note - remember to follow repo conventions for branches, commits and PR titles. Include the Linear ID so your PRs can be linked with Linear tickets.

So branch: `feature/dev-23220-add-anti-aliasing`, commit and PR: "[DEV-23220] Add font anti-aliasing".

When using Claude and Codex, these tools can run `git log` or look at past PRs with the `gh` CLI. We could probably have that in some skill or at least `AGENTS.md`.

PR #1458, opened 2026-06-29
Add hub_layout setting with zeekr-style market dropdown

JesseWynants:
## Context

The Prezly API ships a per-newsroom **`hub_layout`** field with values `tiles` (default) or `market_dropdown`. The `market_dropdown` UI was first shipped in the zeekr theme (hardcoded to zeekr's brand); this PR brings it into the bea base theme, generalized so any client can use it.

**Important:** `hub_layout` is a field on the **Newsroom root** (PATCHed via `admin-api/v1/newsrooms/:id`), **not** a Bea theme setting — the Bea theme-settings schema is `additionalProperties: false`, so the field would be stripped there. The component therefore reads `newsroom.hub_layout`, not `settings.hub_layout`.

**Behavior:**
- `hub_layout = tiles` (default): everything renders as today — hub tile grid + regular language dropdown. No change.
- `hub_layout = market_dropdown`: the language selector becomes a market dropdown built from the hub's sibling/parent newsrooms, and the hub root renders as a regular single-site Stories page (the logo tile grid + aggregated member stories are hidden). Peer sites remain reachable via the dropdown.

## Changes

- **`src/types.ts`** — `NewsroomWithHubLayout` (SDK 26.5.0's `Newsroom` type predates the field; local augmentation).
- **`src/modules/Header/Languages.tsx`** — gate on `newsroom.hub_layout`. Default path unchanged. In `market_dropdown` mode, build `[currentMarket, ...peerMarkets]`: on the hub root peers come from `memberNewsrooms`; on a member site from `newsroom.hub.siblings` + `newsroom.hub.parent` (parent folded in as a normal market). Labels auto-derived from `display_name` with the brand prefix common to the hub's newsrooms stripped at runtime.
- **`src/modules/Header/ui/MarketsPanel/`** (new) — the market dropdown UI (ported from zeekr; no "Other Europe" footer). Current market first + highlighted; cross-site links to `https://{domain}/{locale-slug}`.
- **`app/[localeCode]/(index)/page.tsx`** — render `HubStories` only when `is_hub && hub_layout !== 'market_dropdown'`; otherwise the regular single-site `Stories`.
- **`src/modules/Header/Header.tsx`** — pass the already-fetched `memberNewsrooms` to `<Languages>`.

## Verification
- `tsc --noEmit` and `biome check` pass.
- Ran locally against a hub newsroom with `hub_layout = market_dropdown`: header renders `MarketsPanel` (trigger "{site} / {language}"), the regular dropdown is gone, the hub tile grid is hidden, and the opened dropdown lists the current site first (highlighted) followed by member sites with cross-domain locale links.

## Notes
- Brand-prefix stripping requires every site name — including the hub root — to share a `"Brand "` prefix ending at a word boundary; otherwise the full `display_name` is kept.

### DIFF
--- a/src/modules/Header/Languages.tsx
+++ b/src/modules/Header/Languages.tsx
@@ -3,6 +3,7 @@ import type { Locale } from '@prezly/theme-kit-nextjs';
 
 import { app, routing } from '@/adapters/server';
 import { MarketsPanel, type Market } from '@/modules/Header/ui/MarketsPanel';
+import type { NewsroomWithHubLayout } from '@/types';
 
 import * as ui from './ui';
 
@@ -14,12 +15,11 @@ interface Props {
 }
 
 export async function Languages({ localeCode, memberNewsrooms }: Props) {
-    const newsroom = await app().newsroom();
+    const newsroom = (await app().newsroom()) as NewsroomWithHubLayout;
     const languages = await app().languages();
-    const settings = await app().themeSettings();
     const { generateUrl } = await routing();
 
-    if (settings.hub_layout !== 'market_dropdown') {
+    if (newsroom.hub_layout !== 'market_dropdown') {
         const homepages: ui.Languages.Option[] = languages.map((lang) => ({
             code: lang.code,
             href: generateUrl('index', { localeCode: lang.code }),
--- a/src/modules/Header/Languages.tsx
+++ b/src/modules/Header/Languages.tsx
@@ -1,4 +1,4 @@
-import type { Newsroom } from '@prezly/sdk';
+import { Newsroom } from '@prezly/sdk';
 import type { Locale } from '@prezly/theme-kit-nextjs';
 
 import { app, routing } from '@/adapters/server';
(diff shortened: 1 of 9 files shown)


src/modules/Header/Languages.tsx
```diff
@@ -1,33 +1,160 @@
...
+            />
+        );
+    }
+
+    // market_dropdown: one market per newsroom in the hub, current site first.
+    // Peers come from one of two sources depending on context:
+    // - on the hub root, memberNewsrooms holds the full Newsroom objects (fetched by Header.tsx)
+    // - on a member site, newsroom.hub exposes the lighter HubPeer siblings + parent
+    const peers: Peer[] = newsroom.is_hub
+        ? memberNewsrooms.map(toPeer)
```
fgyimah: Should this filter for only active newsrooms so we don't include newsrooms that are inactive. See https://github.com/prezly/theme-nextjs-bea/blob/920a970a1ce5c37fc03e497aa152e71694c5a465/src/modules/HubStories/HubStories.tsx#L61

src/modules/Header/Languages.tsx
```diff
@@ -1,33 +1,160 @@
...
+    const currentMarket: Market = {
+        countryName: stripPrefix(newsroom.display_name || newsroom.name, sharedPrefix),
+        newsroomUuid: newsroom.uuid,
+        isCurrent: true,
+        languages: languages
+            .filter(
+                (lang) =>
+                    lang.code === localeCode ||
+                    (newsroom.is_hub
+                        ? lang.hub_public_stories_count > 0
```
fgyimah: When `hub_layout === 'market_dropdown'`, the hub root no longer renders `HubStories`, it renders the regular `Stories` module instead.

So I think we should check `lang.public_stories_count` here, not `lang.hub_public_stories_count`.

`hub_public_stories_count` includes stories from member newsrooms too. That can make us show a language even when the root newsroom has no stories in that language. For example:

- root newsroom has `0` German stories
- a member newsroom has `3` German stories
- `lang.hub_public_stories_count` is `3`
- `lang.public_stories_count` is `0`

The dropdown would show German, but clicking it opens the root newsroom German page, which now renders `Stories` and ends up empty.

fgyimah (approved):
Only 2 concerns, everything else looks good

src/modules/Header/Languages.tsx
JesseWynants: Good catch — done in afaac02e. Peers now run through `Newsroom.isActive` (same as `HubStories`), so inactive/deleted members are dropped. Verified on the test hub: it had 3 inactive members (FREE GAMES, GIF Logos, testing-site) which no longer appear in the dropdown.

src/modules/Header/Languages.tsx
JesseWynants: Agreed, fixed in afaac02e. Since the hub root renders the regular `Stories` module in `market_dropdown` mode, the current market now gates on `lang.public_stories_count` (root only) instead of `hub_public_stories_count`, so we won't surface a language whose root homepage would be empty.

PR #1459, opened 2026-06-30
Localize market dropdown current-site name to the viewing locale

JesseWynants:
Follow-up to #1458.

## Problem
In the market dropdown, the current market's label used `newsroom.display_name` — the **default-locale** newsroom name. So on the English version of a site it showed the default name (e.g. "België") instead of the translated one ("Belgium").

## Fix
Use the current locale's `company_information.name` (the same per-locale newsroom name the header uses for the site title via `language.company_information`), falling back to `display_name`/`name`. This name also feeds the shared brand-prefix stripping, so the localized country label comes out clean (e.g. "Belgium" / "België").

```ts
const currentNewsroomName =
    languages.find((lang) => lang.code === localeCode)?.company_information.name ||
    newsroom.display_name ||
    newsroom.name;
```

## Known limitation
**Peer** market names still use their default `display_name` — the hub member/sibling data (`newsroomHub.list` / `newsroom.hub`) doesn't carry per-locale names, so localizing peers would require a separate language fetch per peer. Out of scope here; happy to follow up if we want peers localized too.

## Verification
- `tsc --noEmit` passes.
- Note: couldn't show a visual diff on the test hub since it only has one culture (English); the change mirrors the established `Header.tsx` pattern (`information.name || newsroom.display_name`).

### DIFF
--- a/src/modules/Header/Languages.tsx
+++ b/src/modules/Header/Languages.tsx
@@ -50,15 +50,22 @@ export async function Languages({ localeCode, memberNewsrooms }: Props) {
     // Skip inactive newsrooms — they aren't live, so linking to them is a dead end.
     const peers: Peer[] = candidates.filter((peer) => Newsroom.isActive(peer)).map(toPeer);
 
+    // Use the newsroom name translated to the current locale (company_information.name),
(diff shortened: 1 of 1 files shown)


fgyimah (approved):
👍🏾

PR #1461, opened 2026-07-09
[DEV-23603] Sanitize story and gallery payloads on public surfaces

camilb:
## Summary

Theme-side fix for  the unauthenticated `/api/stories` endpoint (and story-page RSC payloads) returned raw Prezly API objects, exposing internal account metadata on `story.author` (`sign_in_flow`, `last_seen_at`, `locked_until`, `is_terms_of_service_accepted`, `use_case_answer`, `created_at`, numeric user IDs) and newsroom admin `links`.

- New `sanitizeStory.ts` / `sanitizeGallery.ts` utils strip user references (`author`, `pinned_by`, gallery `creator`) down to the display fields the theme actually renders (`avatar_url`, `display_name`, `email`, `first_name`, `last_name`) and drop newsroom `links`.
- Applied in the public API routes: `/api/stories`, `/api/hub-stories`, `/api/galleries`.
- Applied wherever server components pass stories/galleries into client components, so the sanitized shape is what lands in RSC payloads: index/category/tag/hub story lists, related stories, media pages.
- `BroadcastStory` / `BroadcastGallery` contexts narrowed to `{ id, uuid }` — their only consumer (Analytics) uses just the `uuid`.

This is defense-in-depth: the root cause is fixed API-side in the companion PR (prezly/prezly#19042), which covers all themes. This PR ensures the Bea-based builds never republish more than they render, regardless of what the API returns.

## Deployment note

`newsrooms-httpcache` (Varnish) caches theme responses for **7 days** — issue a BAN purge after deploy so cached `/api/stories` responses and story HTML stop serving the old payloads.

## Testing

- `tsc --noEmit` passes across the repo.
- `biome check` clean on all changed files.


JesseWynants:
Nice catch on the extra surfaces — `pinned_by`, gallery `creator`, and especially `newsroom.links` were all leaking alongside `story.author`, and the Varnish BAN note is easy to forget.

I verified the companion API PR (prezly/prezly#19042) locally against a live API and it's clean: metadata gone from `/v2/stories`, `/v1/me` still returns the owner's own fields, zero regressions across the 554-test story suite. I've closed my duplicate (prezly/prezly#19043) and pushed a regression test onto Camil's branch.

Two things about the sanitized shape here, both worth a deliberate decision rather than inheriting them by default:

**1. `sanitizeStory` keeps `email` on the user reference.** That's a CMS login address, not a press contact. The Intigriti report justified it as "author names and emails are intentionally shown on story pages as press contacts" — but that conflates two different objects. Journalist-facing spokespeople are `NewsroomContact` records with their own `email`/`phone`/`company`, reaching story pages via `attached_contacts_content` and the JSON feed. `story.author` is the `sf_guard_user` who clicked publish (`press_release.user_id`), which is exactly why the leaked payload had `sign_in_flow` and `last_seen_at` on it — those columns only exist on `sf_guard_user`.

The triager's escalation path was the *pair*: `sign_in_flow: password` tells you the account has no SSO/MFA, and the email is what you then brute-force. Stripping the former while republishing the latter stops halfway.

Worth checking whether any Bea surface actually renders `author.email` — when I grepped all 21 themes for `story.author` I found zero reads (the only hits were `authorInfo` as a CSS class in flydubai's `StoryCard.tsx` and `class-variance-authority` in help's `button.tsx`). If nothing renders it, it doesn't need to survive sanitization.

**2. The API still sends `newsroom.links` to every other theme.** This PR strips it for Bea, but `/v2/stories` on Camil's branch still returns `links.contacts_management`, `links.domain_settings`, `links.privacy_settings` etc., complete with account slug and site UUID — the `Admin URL:` line from the reporter's own PoC. The ~19 non-Bea themes and any Content API consumer still receive them. Might be worth folding into the API PR rather than fixing per-theme.

Neither is a blocker for this PR — defense-in-depth is the right call regardless of what the API returns.

fgyimah (approved):
Looks good!

PR #1477, opened 2026-07-28
[dev-23729] Match masonry hero image proportions

alpower:
## What changed

- preserve the featured story image's natural aspect ratio when the story list uses masonry layout
- widen the desktop masonry hero to a 60/40 image-to-content split
- keep the existing grid hero and mobile/tablet stacking unchanged
- kept knowledge of layout `layout === 'masonry'` in the parent component where it already was.

## Why

The featured story wrapper was always forced to a 4:3 aspect ratio. Wider thumbnails, including 16:9 images, were therefore cropped even though masonry story cards preserve their source proportions.

## Impact

Masonry newsrooms show the complete featured thumbnail at desktop widths. Grid newsrooms and the full-width featured-story variant retain their current presentation.

Affects 334 live newsrooms.

## Validation

- `biome check` on the changed TypeScript files
- `tsc --noEmit --incremental`
- `git diff --check`

### before
CleanShot 2026-07-28 at 15 54 27@2x

### After
With masonry turned on:
CleanShot 2026-07-28 at 15 54 43@2x
CleanShot 2026-07-28 at 15 50 41@2x

With Masonry turned off (ie Grid):
CleanShot 2026-07-28 at 15 51 46@2x

### DIFF
--- a/src/components/StoryCards/HighlightedStoryCard.tsx
+++ b/src/components/StoryCards/HighlightedStoryCard.tsx
@@ -19,6 +19,7 @@ type Props = {
     fullWidth: boolean;
     newsroomLogo: UploadedImage | null;
     newsroomName: string;
+    preserveImageRatio: boolean;
     rounded: boolean;
     showDate: boolean;
     showSubtitle: boolean;
@@ -29,6 +30,7 @@ export function HighlightedStoryCard({
     fullWidth,
     newsroomLogo,
     newsroomName,
+    preserveImageRatio,
     rounded,
     showDate,
     showSubtitle,
@@ -111,6 +113,7 @@ export function HighlightedStoryCard({
             fallback={fallback}
             layout="horizontal"
             placeholder={{}}
+            preserveImageRatio={preserveImageRatio}
             publishedAt={story.published_at}
             showDate={showDate}
             showSubtitle={showSubtitle}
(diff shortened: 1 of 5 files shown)


harmasz (approved):
Approved — the masonry hero now preserves the natural thumbnail aspect ratio while leaving the grid and full-width variants unchanged.

Non-blocking follow-up: the new 60/40 split makes the desktop hero image roughly 690px wide at the maximum container size, while StoryImage still advertises the hero size as 575px. Consider adding a dedicated size hint for this variant so browsers do not select a slightly undersized srcset candidate.

christianprezly (approved):
small edge case, not worth blocking: if the featured story has no thumbnail, the fallback placeholder now gets .static → aspect-ratio: 1.5 instead of the wrapper's 4:3 — a minor visual shift, but again consistent with masonry cards.

Tiny nit — withStaticImage || preserveImageRatio is computed twice in StoryCard.tsx; could be hoisted to a const for readability.

alpower:
> Non-blocking follow-up: the new 60/40 split makes the desktop hero image roughly 690px wide at the maximum container size, while StoryImage still advertises the hero size as 575px. Consider adding a dedicated size hint for this variant so browsers do not select a slightly undersized srcset candidate.

Fixed this.

alpower:
> Tiny nit — withStaticImage || preserveImageRatio is computed twice in StoryCard.tsx; could be hoisted to a const for readability.

hoisted the const Thanks. I'm not bothered by fallback, as new 3:2 fallback is consistent with the other masonry cards

PR #1482, opened 2026-08-13
DEV-23550: Remove the mask in story in-app previews

boki-prezly:
### Summary

- Replace the full-page PreviewPageMask overlay with a ?mask=true navigation lock that only blocks `<a href/>`  clicks/auxclicks. 
- Keeps in-app preview embeds interactive (YouTube, galleries, PDFs) while still preventing redirects away from the previewed story.

[screen-capture (6).webm](https://github.com/user-attachments/assets/8604d024-e89f-4866-a857-e0e977e1a114)

### DIFF
--- a/src/components/PreviewPageMask/PreviewPageMask.module.scss
+++ b/src/components/PreviewPageMask/PreviewPageMask.module.scss
@@ -1,6 +1,0 @@
-.mask {
-    position: fixed;
-    inset: 0;
-    background: transparent;
-    z-index: 9999;
-}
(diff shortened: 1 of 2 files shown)


harmasz (approved):
Looks good! One thing to keep in mind: this blocks normal link clicks, but forms can still navigate away. For example, pressing Enter in the search box can replace the preview with the search page, and submitting the email form can take it to another page. It might be worth blocking form-driven navigation too, while still allowing things like videos and image galleries to work.

fgyimah (approved):
👍🏾

PR #1483, opened 2026-08-20
Fix wrapped story image sizing 

zawropati:
## Summary

Previously, the editor allowed choosing sizes for wrapped images, but the Bea theme renderer was overriding those values.
This PR makes the theme respect the wrapped image size:
- small renders at 200px
- best-fit renders at up to 360px, capped at half of the story row width

It also adjusts the wrapped image spacing in the theme, since the editor spacing does not apply to the rendered newsroom story.

Small:
CleanShot 2026-08-20 at 10 09 18@2x

Best-fit:
CleanShot 2026-08-20 at 10 34 50@2x

### DIFF
--- a/src/components/ContentRenderer/ContentRenderer.module.scss
+++ b/src/components/ContentRenderer/ContentRenderer.module.scss
@@ -1,5 +1,4 @@
 $avatar-size: 60px;
-$wrapped-image-width: 320px;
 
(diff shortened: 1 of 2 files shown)


fgyimah (approved):
Digital art gif. A giant green hand with legs smiles happily and bounces, giving a big thumbs up.

